PR #47, opened 2018-01-18
Crashlytics

tobiasrohloff:
Enable Firebase Crashlytics and use `git-crypt` for protecting non-public files.


tobiasrohloff:
@mathebox for your reference.

PR #71, opened 2018-05-05
Fixes for Channel list and details

bensoftde:
Fixes:
- Course details instead of learning pane were opened upon click in Channel list slider although user was enrolled for this Course
- misleading "+" button at end of Channel list slider now replaced by card saying "show more"
- upon clicking this card, the toolbar is now collapsed and the screen is being scrolled to the element that would be next in the slider

screen

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.java
+++ b/app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.java
@@ -42,7 +42,8 @@ public class ChannelDetailsFragment extends LoadingStatePresenterFragment<Channe
 
     @AutoBundleField String channelId;
 
-    @AutoBundleField (required = false) boolean scrollToCourses = false;
+    //-1 do not scroll to Course
+    @AutoBundleField(required = false) int scrollToCoursePosition = -1;
 
     @BindView(R.id.layout_header) FrameLayout layoutHeader;
     @BindView(R.id.image_channel) ImageView imageChannel;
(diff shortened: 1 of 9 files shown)


app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.java
```diff
@@ -134,23 +135,29 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
             courseListAdapter.update(courses);
 
-        if(scrollToCourses)
-            courseList.scrollToPosition(1);
+        if (scrollToCoursePosition >= 0) {
+            try {
+                ChannelDetailsActivity a = ((ChannelDetailsActivity) getActivity());
+                if (a != null)
+                    a.appBarLayout.setExpanded(false);
+                courseList.smoothScrollToPosition(scrollToCoursePosition + 3);
```
tobiasrohloff: What does the magic number `3` here?

bensoftde: description + heading + (n+1)th element (as the first element is 1 and not 0)

bensoftde:
Card now cut off and text changed to "show more courses"

screen

app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.java
tobiasrohloff: Would be a good comment on top of this line :)

Issue #76, opened 2018-06-07
Try Architecture Components with new Feature


tobiasrohloff:
Let's pick a new feature and build it with architecture components to identify what needs to be done and changed in our current architecture.

PR #85, opened 2018-08-01
Documents with Architecture Components

### LINKED ISSUE #57: Documents
Add documents feature for OpenWHO.

tobiasrohloff:
Fixes #57 #76 

First of all, sorry for this huge PR. Too much stuff, I know.

This PR mainly introduces the Documents feature for OpenWHO and uses the Architecture Components for it (ViewModel and LiveData), instead of the former MVP approach. `NetworkStateFragment` and `NetworkJob` are the most interesting base classes, besides everything in the `viewmodels` package. Other changes include:

*  Adjusted the initial storage migration. All course item downloads are now under `/Courses` and all documents under `/Documents` (a document can be assigned to multiple courses and maybe items in the future). Since the SD Card feature is not released yet, I was able to use the existing migration.
*  Added Documents to the `DownloadsFragment` overview and cleaned the code a bit
*  The `CourseActivity` tabs are now configured by the `CourseArea` enum class with better support for platform-dependent features
*  Generic `DownloadViewHelper`
*  Moved the `RealmSchemaMigration` class into an own package
*  Material Design 2 button and card styles
*  A lot of code migrated to Kotlin


tobiasrohloff:
@bensoftde please test the storage migration and download with your SD card device.

app/src/main/java/de/xikolo/controllers/downloads/DownloadsFragment.kt
```diff
@@ -0,0 +1,253 @@
...
+        val TAG: String = DownloadsFragment::class.java.simpleName
+    }
+
+    private var adapter: DownloadsAdapter? = null
+
+    private var downloadManager: DownloadManager? = null
+
+    private var permissionManager: PermissionManager? = null
+
+    private var notificationController: LoadingStateHelper? = null
```
bensoftde: Can't we do these as lateinit to avoid nullability?

app/src/main/java/de/xikolo/controllers/downloads/DownloadsFragment.kt
```diff
@@ -0,0 +1,253 @@
...
+import java.io.File
+import java.util.*
+
+class DownloadsFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener, DownloadsAdapter.OnDeleteButtonClickedListener {
+
+    companion object {
+        val TAG: String = DownloadsFragment::class.java.simpleName
+    }
+
+    private var adapter: DownloadsAdapter? = null
```
bensoftde: Can't we do these as `lateinit` to avoid nullability?

tobiasrohloff: They all kinda depend on `activity`, which is marked as nullable now.

PR #90, opened 2018-08-27
Update video player to ExoPlayer2

bensoftde:
Updates the used video player to [ExoPlayer2](https://google.github.io/ExoPlayer/).
To keep changes to a minimum, the `ExoPlayerVideoView` functions as a wrapper using the same function and interface names as the previous one. It can be easily extended to e.g. support HLS videos in the future.

### DIFF
--- a/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
+++ b/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
@@ -70,7 +70,6 @@ open class ExoPlayerVideoView : PlayerView {
 
                 if (playbackState == Player.STATE_ENDED) {
                     onCompletionListener?.onCompletion()
-                    player.seekTo(0)
                 }
             }
 
--- a/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
+++ b/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
@@ -19,11 +19,11 @@ import com.google.android.exoplayer2.util.Util
 
 open class ExoPlayerVideoView : PlayerView {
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
```diff
@@ -0,0 +1,162 @@
...
+import com.google.android.exoplayer2.upstream.DefaultDataSourceFactory
+import com.google.android.exoplayer2.util.Util
+
+open class ExoPlayerVideoView : PlayerView {
+
+    private lateinit var mContext: Context
+    private lateinit var mPlayer: SimpleExoPlayer
+    private lateinit var mPlayerListener: Player.EventListener
+    private val mBandwidthMeter: DefaultBandwidthMeter = DefaultBandwidthMeter()
+    private var mMediaSource: MediaSource? = null
```
tobiasrohloff: [Just Say mNo to Hungarian Notation](https://jakewharton.com/just-say-no-to-hungarian-notation/)

app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
```diff
@@ -0,0 +1,162 @@
...
+
+    fun seekTo(position: Long) {
+        player.seekTo(position)
+    }
+
+    fun setPlaybackSpeed(speed: Float) {
+        player.playbackParameters = PlaybackParameters(
+            speed,
+            player.playbackParameters.pitch,
+            player.playbackParameters.skipSilence)
```
tobiasrohloff: The closing parenthesis should be moved to the next line.

PR #93, opened 2018-09-21
Video Player Settings

### LINKED ISSUE #38: Subtitles
When subtitles are available for a video, the user should be able to display them. This is supported since Android 4.4: https://www.youtube.com/watch?v=hCRGc2PcmB8

Therefore, the new API has to deliver links to the `vtt` files.

bensoftde:
Reorganizes video player settings.

Major changes:
- The UI now shows a settings icon instead of single settings items
- Settings states are moved out of `VideoHelper` into `VideoSettingsHelper` which is also responsible for creating the settings layouts

EDIT: Now also shows `(offline)` behind a quality when it is available offline

Fixes #38 

settings


app/src/main/java/de/xikolo/controllers/helper/VideoSettingsHelper.kt
```diff
@@ -0,0 +1,243 @@
...
+        )
+        list.addView(
+            buildSettingsItem(
+                R.string.icon_speed,
+                context.getString(R.string.video_settings_speed) + "  " + context.getString(R.string.video_settings_separator) + "  " + currentSpeed.toString(),
+                View.OnClickListener { clickListener.onPlaybackSpeedClick() },
+                false
+            )
+        )
+        if (subtitles != null && subtitles.isNotEmpty()) {
```
tobiasrohloff: Could be simplified: `subtitles?.isNotEmpty() == true`

app/src/main/java/de/xikolo/controllers/helper/VideoSettingsHelper.kt
```diff
@@ -0,0 +1,243 @@
...
+    var currentVideoSubtitles: VideoSubtitles? = null
+
+    fun buildSettingsView(): ViewGroup {
+        val list = buildSettingsPanel(null)
+
+        list.addView(
+            buildSettingsItem(
+                R.string.icon_quality,
+                context.getString(R.string.video_settings_quality) + "  " + context.getString(R.string.video_settings_separator) + "  " + currentQuality.toString() +
+                    if (qualityOfflineInfo.isOfflineAvailable(currentQuality)) " " + context.getString(R.string.video_settings_quality_offline) else "",
```
tobiasrohloff: Is the `else` required?

app/src/main/java/de/xikolo/controllers/helper/VideoSettingsHelper.kt
```diff
@@ -0,0 +1,243 @@
...
+                false
+            )
+        )
+        if (subtitles != null && subtitles.isNotEmpty()) {
+            list.addView(
+                buildSettingsItem(
+                    R.string.icon_subtitles,
+                    context.getString(R.string.video_settings_subtitles) + if (currentVideoSubtitles != null)
+                        "  " + context.getString(R.string.video_settings_separator) + "  " + Locale(currentVideoSubtitles?.language).displayLanguage
+                    else
```
tobiasrohloff: Is the else required?

app/src/main/java/de/xikolo/controllers/helper/VideoSettingsHelper.kt
bensoftde: Sadly yes, because `if` requires the `else` branch if used in an expression: [see here](http://kotlinlang.org/docs/reference/control-flow.html#if-expression)

PR #94, opened 2018-09-22
Video Seekbar Preview

### LINKED ISSUE #81: PreviewSeekBar for Videos
Let's check if we can integrate https://github.com/rubensousa/PreviewSeekBar for our videos.

bensoftde:
Adds a video preview image above the seek bar.

This works for videos saved locally as well as streamed ones. Under the hood, the framework uses a native method, so I assume it's hardware accelerated. When grabbing preview images over the network, the framework manages to only download the relevant parts of the video.
Preview generation takes approximately 100-200ms locally and around 1500ms streamed (16Mbps connection).
To minimize the number of loaded previews, a preview is only generated when the interval between previews is higher than 100ms and the user seeked for at least 5 seconds in either direction on the seekbar.

Notice: This PR depends on #93 

Fixes #81 

previewseekbar

### DIFF
--- a/app/dependencies.gradle
+++ b/app/dependencies.gradle
@@ -46,6 +46,7 @@ dependencies {
 
     // video
     implementation 'com.google.android.exoplayer:exoplayer:2.9.0'
+    implementation 'com.github.rubensousa:previewseekbar:2.0.0'
 
     // view binding
     def butterknifeVersion = '8.8.1'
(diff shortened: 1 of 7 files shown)


tobiasrohloff:
@bensoftde I changed the base to `master` now since #93 was merged. Please take care of the conflicts.

PR #96, opened 2018-10-05
Death of Android 4

tobiasrohloff:
Remove all Android 4 related parts.

Please review carefully and test on emulator and device.


tobiasrohloff:
@bensoftde good catch! I refactored that.

PR #115, opened 2019-01-30
Migration of some course specific things to ViewModel architecture

bengelhaupt:
Introduces ViewModel architecture for the following components:
- Main Course List
- Main Certificate List
- Course Description
- Course Certificates

Added
- `CourseListViewModel` for when having to observe multiple courses
- `CourseViewModel` for when having to observe a single course`s properties

Some basic course functionality might not be migrated to ViewModel architecture as it is still used in `Channels`, which is not yet migrated, thus having some code duplications.

PR #116, opened 2019-01-30
Announcements ViewModel Refactorings

bengelhaupt:
Merges the `ListGlobalAnnouncementsJob` and `ListCourseAnnouncementsJob`.

Renames `AnnouncementsViewModel` into `AnnouncementListViewModel` and `DocumentsViewModel` into `DocumentListViewModel` respectively.

Also improves handling of observed data.


bengelhaupt:
> I think they should be named `AnnouncementListViewModel` and `DocumentListViewModel`. Sooner or later this will lead to problems. Think about the two ViewModels you introduce here #115.

I renamed the ViewModels and this branch could now be merged.

tobiasrohloff:
Thank you :)

PR #117, opened 2019-02-04
Migrate to AndroidX

### LINKED ISSUE #110: Migrating to AndroidX
https://developer.android.com/jetpack/androidx/

tobiasrohloff:
This migrates the whole project to AndroidX (and Android Studio 3.3).

Fix #110


app/src/main/java/de/xikolo/controllers/helper/LoadingStateHelper.java
```diff
@@ -22,7 +22,8 @@
 
     @BindView(R.id.container_content_message) FrameLayout messageContainer;
     @BindView(R.id.progress_bar) ProgressBar progressBar;
-    @BindView(R.id.refresh_layout) SwipeRefreshLayout refreshLayout;
+    @BindView(R.id.refresh_layout)
+    SwipeRefreshLayout refreshLayout;
```
bengelhaupt: Shouldn't this be formatted like the other lines?

PR #118, opened 2019-02-19
Testing Base

bengelhaupt:
Introduces first testing experiments and testing base framework.

Has UI tests for `AppStartup`, `Channels`, `Login` and `Navigation`.
The api requests for these tests are being intercepted and the responses mocked by returning contents of `.json` files. All of the new classes are documented.

NOTICE: The `.travis.yml` has been changed to allow CI testing on emulator but is still very very slow and emulators might not start up completely in the 10-minute-timeout which makes the CI build fail. This is because Travis' servers do not support x86 emulators as of now and ARM emulators are being used.

-----------------------
What should be changed in this PR to have a working testing base?
- remove emulators from `.travis.yml` for now to have working CI tests?
- break mocked data further down (e.g. only having a few fake courses/channels)? What about non-api resources like images and videos?
- is `ApiService` integration like this okay?


app/src/androidTest/java/de/xikolo/mocking/MockingInterceptor.kt
```diff
@@ -0,0 +1,36 @@
...
+class MockingInterceptor(private val context: Context) : Interceptor {
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return try {
+            val mockedResponse = MockingData.getResponse(context, chain.request(), BuildConfig.X_FLAVOR)!!
+            Response.Builder()
+                .code(mockedResponse.statusCode)
+                .message(mockedResponse.responseString)
+                .request(chain.request())
+                .protocol(Protocol.HTTP_1_0)
```
tobiasrohloff: I guess this should be at least http1.1

tobiasrohloff:
* remove emulators from `.travis.yml` for now to have working CI tests?
  * yes, if travis doesn't work sufficiently we will use local testing
* break mocked data further down (e.g. only having a few fake courses/channels)? What about non-api resources like images and videos?
  * ignore such media files for now
  * we don't need 200 courses, something like 10 should be ok

app/src/androidTest/java/de/xikolo/ui/AppStartupTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.xikolo.ui.helper.NavigationHelper.Companion.WAIT_STARTUP
+import org.junit.Rule
+import org.junit.Test
+
+@LargeTest
+class AppStartupTest : BaseMockedTest() {
+
+    @Rule
+    @JvmField
+    var mActivityTestRule = ActivityTestRule(SplashActivity::class.java)
```
tobiasrohloff: No hungarian notation pls

app/src/androidTest/java/de/xikolo/unit/DownloadManagerTest.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import androidx.test.filters.LargeTest
+import de.xikolo.managers.DownloadManager
+import org.junit.Before
+
+/**
+ * This is about to become a unit test for the {@link DownloadManager} class.
+ */
+@LargeTest
+class DownloadManagerTest {
```
tobiasrohloff: Please remove if empty

bengelhaupt:
I could remove some of the waiting statements because Espresso somehow (I guess via the Activity thread) determines when loading is "finished". But this does not work everytime, e.g. when logging in you have to wait until the network processing is finished.

tobiasrohloff:
@bengelhaupt I generalized the `ApiService` manipulation. Still not perfect, but works for now. Please do a final review and test.

PR #121, opened 2019-02-28
ViewModel architecture for course learnings

bengelhaupt:
Course Learnings are now migrated to ViewModel architecture.

Refactored some of the Adapters to Kotlin too.


app/src/main/java/de/xikolo/models/dao/course/LearningsDao.kt
```diff
@@ -0,0 +1,29 @@
+package de.xikolo.models.dao.course
+
+import androidx.lifecycle.LiveData
+import de.xikolo.models.Item
+import de.xikolo.models.dao.base.BaseDao
+import de.xikolo.viewmodels.base.asLiveData
+import io.realm.Realm
+import io.realm.kotlin.where
+
+class LearningsDao(realm: Realm) : BaseDao(realm) {
```
tobiasrohloff: This should be `ItemsDao`.

Issue #109, opened 2018-12-04
Refactor Kotlin-Realm-Extensions


tobiasrohloff:
We use `com.github.vicpinm:krealmextensions`, but only very few of them. Could be replaced with some own extensions.

PR #128, opened 2019-03-26
Refactor DAOs and Models

tobiasrohloff:
This PR migrates all database accesses to the DAOs and introduces some conventions for them:
- single entity requests begin with `find`, collection requests begin with `all`
- managed requests with results wrapped as `LiveData` are implemented as instance methods of each DAO
  - some sugar for default `all` and `find` cases is implemented by the `BaseDao`
- unmanaged requests with plain object/list results are available via an `Unmanaged` object in each DAO
  - requests are class methods, since no managed realm instance and therefore no state is needed
  - a new `#asCopy` extensions must be called to ensure no Realm artifacts are used outside the DAOs
- some model classes still have convenient accessor methods to return relationship object
  - these are always unmanaged and only call the corresponding DAO method

In the future, I will try to further enforce some of these conventions, but at least all the code lives inside the DAO classes now and we have a standartized way of accessing managed and unmanaged data.

This PR also fixes https://github.com/openHPI/xikolo-android/issues/109.


app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsActivity.kt
```diff
@@ -0,0 +1,112 @@
...
+    lateinit var imageView: ImageView
+
+    @BindView(R.id.appbar)
+    lateinit var appBarLayout: AppBarLayout
+
+    @BindView(R.id.collapsing_toolbar)
+    lateinit var collapsingToolbar: CollapsingToolbarLayout
+
+    @BindView(R.id.stub_bottom)
+    lateinit var stubBottom: ViewStub
```
bengelhaupt: This variable is unused, should it be removed?

app/src/main/java/de/xikolo/models/dao/CourseDao.kt
```diff
@@ -0,0 +1,202 @@
...
+                            EnrollmentDao.Unmanaged.findForCourse(course.id)
+                                ?.anyCertificateAchieved() == true
+                        }
+                }
+
+            fun search(query: String?, withEnrollment: Boolean): List<Course> =
+                Realm.getDefaultInstance().use { realm ->
+                    val realmQuery = realm.where<Course>()
+                        .equalTo("external", false)
+                        .beginGroup()
```
bengelhaupt: group indentation

app/src/main/java/de/xikolo/models/Video.java
```diff
@@ -44,14 +44,6 @@
...
 
-    public static Video get(String id) {
-        Realm realm = Realm.getDefaultInstance();
-        Video model = realm.where(Video.class).equalTo("id", id).findFirst();
-        if (model != null) model = realm.copyFromRealm(model);
-        realm.close();
-        return model;
-    }
-
     public static Video getForContentId(String contentId) {
```
bengelhaupt: why is this left over?

app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsActivity.kt
tobiasrohloff: Done.

app/src/main/java/de/xikolo/models/dao/CourseDao.kt
tobiasrohloff: Fixed.

app/src/main/java/de/xikolo/models/Video.java
tobiasrohloff: Removed.

tobiasrohloff:
I added some docs to the `BaseDao`.

PR #138, opened 2019-04-05
Fix for ViewModel composition

### LINKED ISSUE #136: Reusage of ViewModel breaks NetworkStateLiveData
The reusage of ViewModels seems to break NetworkStateLiveData updates. For example the `DescriptionViewModel` and `CertificateListViewModel` use 
```
private val courseViewModel = CourseViewModel(courseId)

val course: LiveData<Course> = courseViewModel.course
```
If you refresh the page, the requests succeds but the refresh animation never finishes. I think you also have to delegate the `networkState` to the reused ViewModel.

bengelhaupt:
Fixes #136

### DIFF
--- a/app/src/main/java/de/xikolo/viewmodels/course/CourseViewModel.kt
+++ b/app/src/main/java/de/xikolo/viewmodels/course/CourseViewModel.kt
@@ -22,7 +22,8 @@ class CourseViewModel(val courseId: String) : BaseViewModel() {
         requestCourse(true)
     }
 
-    private fun requestCourse(userRequest: Boolean) {
+    fun requestCourse(userRequest: Boolean) {
         GetCourseJob(courseId, networkState, userRequest).run()
     }
+
 }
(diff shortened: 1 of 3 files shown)


bengelhaupt:
I thought about something like this. But this would mean, that the constructors of reused ViewModel's would need to change like this: `class CourseListViewModel(private val filter: CourseListFilter, private val parentNetworkState: NetworkStateLiveData = null) : BaseViewModel()` because they somehow need to get the `NetworkState` of the "parent" ViewModel they are being used by an then override the `networkState` field (with an additional `null` check if it has no "parent" ViewModel and is being used on its own).
I decided against this this approach as we now would have declaration inconsistencies across all our ViewModels as some are being reused and others not.
I realized that the delegate `request*`-calls broke the network state handling so I decided to only have reused ViewModels expose such calls with an additional `networkState` argument and keep the other call private:
```
private fun request(userRequest: Boolean) {
        Job(networkState, userRequest).run() // ViewModel's NetworkState
}

fun request(userRequest: Boolean, networkState: NetworkStateLiveData) {
        Job(networkState, userRequest).run() // delegatee NetworkState
}
```

bengelhaupt:
Made the proposed changes and some other improvements. Ready for review.

PR #148, opened 2019-06-16
UTC information for dates

### LINKED ISSUE #80: Display Timezone for Dates

bengelhaupt:
Adds a `DateTextView` which opens a `DateInfoDialog` showing the UTC time in addition to local time of a date.

`DateTextView` is a `TextView` having the functions `setDate(...)` and `setDateSpan(...)` that dash-underlines its text and becomes clickable when one of them has been called. Upon click, the `DateInfoDialog` is opened showing a long format of the date (or date span) in the local and UTC timezone. The title of this dialog can be customized with the `res:infoTitle` xml attribute on `DateTextView`. 

The functionality was added to following places in the app:
- Course > Course Details (only when course end date is in future)
- Course > Learnings (when a section is unavailable)
- Course > Course Dates
- Announcements

   

 

Fixes #80

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/announcement/AnnouncementFragment.kt
+++ b/app/src/main/java/de/xikolo/controllers/announcement/AnnouncementFragment.kt
@@ -17,6 +17,7 @@ import de.xikolo.managers.UserManager
 import de.xikolo.models.Announcement
 import de.xikolo.utils.MarkdownUtil
 import de.xikolo.viewmodels.announcement.AnnouncementViewModel
+import de.xikolo.views.DateTextView
 import java.text.DateFormat
 import java.util.*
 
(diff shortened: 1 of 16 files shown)


tobiasrohloff:
Nice one. :) Some picky remarks: Can you use a dotted line instead of a dashed? And for the locked sections the color of the line is darker than the font and also there is a strange padding on the right.

bengelhaupt:
I changed the dashes to dots and made them slightly larger.
Also the underline color is set from the text color.

tobiasrohloff:
Rebased and minor styling adjustments.

tobiasrohloff:
The DateTextView is not used for the Course Date Overview Activity/Fragment. Is that on purpose or has it been forgotten?

bengelhaupt:
> The DateTextView is not used for the Course Date Overview Activity/Fragment. Is that on purpose or has it been forgotten?

That's on purpose as item clicks in this list lead to the respecting course. It could be underlined but clicks would not work.

tobiasrohloff:
I think that's a matter of the click event consumption hierarchy.

bengelhaupt:
Yes it is, but as a user I would find this very misleading to have two overlapping "touch areas" with diffierent semantics in the same item.

tobiasrohloff:
Good point.

PR #142, opened 2019-05-17
Rewritten Video Player Architecture and Course Teaser Videos

bengelhaupt:
_Introduces a Fragment-based video playing architecture to support non-item video's `VideoStream`s._

The new architecture mainly helps in separating functionality specific to item-based videos like downloads or Lanalytics logging as this functionality cannot be provided with plain `VideoStream`s.
This means the `VideoHelper` has been refactored to a `VideoStreamPlayerFragment` which implements this basic video playing functionaliy and a `VideoItemPlayerFragment` which holds the item-specific functionality. As these Fragments maintain their own lifecycle, no external calls from a controller Activity are necessary anymore to e.g. stop the playback on exit. This controller Activity (here `VideoStreamPlayerActivity` and its counterpart `VideoItemPlayerActivity`) can implement a `VideoStreamPlayerFragment.ControllerInterface` to control certain behavior like settings inflation, but do not have to. This makes it possible to inflate these Fragments easily into any layout without major modifications to the Activity.
The `VideoActivity` has been replaced by the `VideoItemPlayerActivity` which now inflates two Fragments, the `VideoItemPlayerFragment` and `VideoDescriptionFragment`.

In addition to these architectural changes, the video player user interface behavior has been simplified and the layout has been reworked to now better support longer-than-16:9-devices and devices that ship with a notch. The user can now choose in the settings or use a pinch gesture to select whether the video content shall extend across the whole screen by extending under the notch and/or scaling the video to fill the whole screen area (sadly the screenshots did not capture the notch, but you can imagine its existence on the left side of the screen). This function is not avialable if the device has a screen in the video's aspect ratio and no notch.

 

And to address the issue all this effort was made for, now a course teaser video can be played from the course details page.

**When deploying please do a staged rollout so eventually occuring problems can be addressed earlier.**

Fixes #61 
Fixes #133 

EDIT: Also Fixes #145

PR #149, opened 2019-06-16
ViewModel composition restructurings

bengelhaupt:
Feature #142 introduced another bug related to ViewModel composition. Occasionally under `Course > Course Details` the teaser video disappeared in favor of the normal course image even though the teaser video existed. This was because of conflicting network requests overriding the states of each other.
That is caused by the fact that the `DescriptionViewModel` depends on the `CourseViewModel` which itself depends on the `CourseListViewModel` and `DateListViewModel` with `DateListViewModel` again depending on `CourseListViewModel`.

To avoid confusion in the future and to clean up these cyclic dependency structures allowing for a less error-prone development process, this draft aims for explicitly having classes used for delegating foreign behavior of ViewModels. These Delegates are located in the `de.xikolo.viewmodels.shared` package and are
- `CourseListViewModelDelegate`
- `CourseViewModelDelegate`
- `AnnouncementListViewModelDelegate`
- `DateListViewModelDelegate`

Because of the fact, that network requesting functions are now also placed there, overriding the `NetworkState` of delgates has become obsolete as the own `NetworkState` has to be supplied as a parameter for those network requests.

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/course/LearningsFragment.kt
+++ b/app/src/main/java/de/xikolo/controllers/course/LearningsFragment.kt
@@ -68,11 +68,6 @@ class LearningsFragment : NetworkStateFragment<LearningsViewModel>(), SectionLis
             }
         ))
 
-        viewModel.course
-            .observe(this) {
-                activity?.title = it.title
-            }
-
         viewModel.accessibleItems
             .observe(this) {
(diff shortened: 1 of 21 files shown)


tobiasrohloff:
I've seen something like this coming 🙈

How about omitting the `ViewModel` from the `Delegate`'s name? I guess it should be clear without as well.

PR #151, opened 2019-06-28
New Navigation Drawer and refactorings from NavigationPresenter to Architecture Components

bengelhaupt:
Introduces a new navigation drawer seen in `MainActivity` and refactored `NavigationPresenter` to Architecture Components.

The navigation is now inflated from a menu file. This allows for a more robust selection of navigation elements beacuse now only the item's id has to be supplied instead of the position (which differed from flavor to flavor). Dynamically changing the menu is also much easier, because altering the `menu` object of the `NavigationView` and invalidating it, is sufficient to change the drawer items.
Because of these simplifications, the `NavigationDrawerFragment` and `NavigationDrawerAdapter` could be removed. This has the tradeoff, that icon resources have to be generated for the menu items.

The pictures below show the drawer for all flavors when logged in and when logged out (only openSAP).

&nbsp;&nbsp;
  

Fixes #131

PR #152, opened 2019-06-29
Refactoring of Login and Profile to Architecture Components

### LINKED ISSUE #130: Migrate LoginPresenter to Architecture Components

### LINKED ISSUE #132: Migrate ProfilePresenter to Architecture Components

bengelhaupt:
Refactors the `LoginPresenter` and `ProfilePresenter` to Architecture Components. No UI changes were made.

Please test carefully as the old version contained nested `RequestJobs`, which made two separate `NetworkState`s necessary (in `LoginViewModel`).

Also, something seems wrong with our `EventBus`. For me methods subscribed for login/ logout events are only called if they are posted sticky with `postSticky(...)`. I had the same issue in #151 . It may be caused by the fact that the receiver is in the background when the event is posted and thus the receiver is not receiving it. This can cause the bug, that upon SSO login, the `SsoLoginActivity` is not finishing when the login succeeds. It is still strange why it works sometimes and sometimes does not.

Fixes #130 
Fixes #132 

_Note that adjustments need to be made if merged after #151_


app/src/main/java/de/xikolo/controllers/login/LoginFragment.kt
```diff
@@ -0,0 +1,267 @@
...
+        if (isEmailValid(email)) {
+            if (password != "") {
+                showProgressDialog()
+                viewModel.login(email, password)
+            } else {
+                editTextPassword.error = getString(R.string.error_password)
+            }
+        } else {
+            editTextEmail.error = getString(R.string.error_email)
+        }
```
tobiasrohloff: Such nested ifs can be simplified with guard statements. For example:
```kotlin
if (!isEmailValid(email)) {
    editTextEmail.error = getString(R.string.error_email)
    return
}

if (password == "") {
    editTextPassword.error = getString(R.string.error_password)
    return
}

showProgressDialog()
viewModel.login(email, password)
```

app/src/main/java/de/xikolo/models/dao/UserDao.kt
```diff
@@ -1,13 +1,21 @@
...
+import de.xikolo.extensions.asLiveData
+import de.xikolo.managers.UserManager
 import de.xikolo.models.User
 import de.xikolo.models.dao.base.BaseDao
 import io.realm.Realm
 import io.realm.kotlin.where
 
 class UserDao(realm: Realm) : BaseDao<User>(User::class, realm) {
 
+    fun getCurrent(): LiveData<User> = query()
```
tobiasrohloff: Either `findCurrent` or simply `current`.

app/src/main/java/de/xikolo/viewmodels/main/ProfileViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+        get() = enrollmentDelegate.enrollmentCount
+
+    override fun onFirstCreate() {
+        userDelegate.requestUserWithProfile(networkState, false)
+        enrollmentDelegate.requestEnrollmentList(networkState, false)
+    }
+
+    override fun onRefresh() {
+        userDelegate.requestUserWithProfile(networkState, true)
+        enrollmentDelegate.requestEnrollmentList(networkState, true)
```
tobiasrohloff: I think the enrollment request network state should not be connected with the UI, it is a secondary resource and requested concurrently with the profile.

app/src/main/java/de/xikolo/viewmodels/login/LoginViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+import de.xikolo.viewmodels.shared.UserDelegate
+
+class LoginViewModel : BaseViewModel() {
+
+    private val userDelegate = UserDelegate(realm)
+
+    val loginNetworkState: NetworkStateLiveData = networkState
+    val profileNetworkState: NetworkStateLiveData by lazy {
+        NetworkStateLiveData()
+    }
```
tobiasrohloff: I think in this case the network state object can be shared by the two sequential requests. Currently, the `profileNetworkState` is useless since it is not connected to the UI. Passing null would do the same.

tobiasrohloff:
Does migrating all login/ logout events to sticky events solves the issue or still flaky behavior? Any downsides of this fix?

app/src/main/java/de/xikolo/models/dao/UserDao.kt
bengelhaupt: Renamed to `current`

app/src/main/java/de/xikolo/viewmodels/main/ProfileViewModel.kt
bengelhaupt: Solved

app/src/main/java/de/xikolo/viewmodels/login/LoginViewModel.kt
bengelhaupt: Solved

bengelhaupt:
Incorporated feedback from discussions. The `EventBus` posting has been reverted to the old behavior.

tobiasrohloff:
Do you still face issues with the SSO?
> the SsoLoginActivity is not finishing when the login succeeds
But in this case, the activity is in the foreground and should receive the event?

bengelhaupt:
> But in this case, the activity is in the foreground and should receive the event?

Actually the `WebViewFragment` inside the `SsoLoginActivity` has a method `interceptSSOLogin(...)` which then opens the `LoginActivity` with a token as parameter to load the user profile. During this time, the `SsoLoginActivity` is in the background. It is intended, that the `LoginEvent` sent from the `LoginActivity` after the profile has been loaded, closes the `SsoLoginActivity`, but that does not work.

bengelhaupt:
Weird thing. What fixed it for me was calling `EventBus.getDefault().post(LoginEvent())` before `activity?.finish()` in the `profileNetworkState`'s success callback in `LoginActivity`. This now closes the `SsoLoginActivity` even though the `LoginActivity` is not yet finished when the event is posted. In the old code, `getVieworThrow().finish()` is called before `EventBus.getDefault().post(LoginEvent())`.

tobiasrohloff:
@bengelhaupt we have some conflicts here...

bengelhaupt:
Merge conflicts have been resolved

PR #154, opened 2019-07-08
App startup enhancements

bengelhaupt:
When opening the app and closing it right away, the window goes away but when the health job finishes, it opens again. This fix keeps track whether the app is visible and only starts the `MainActivity` if the app has not been closed.

Also, is there a way to allow the user to use the app in offline mode, when the network is really slow (like EDGE) and the health job times out with `ERROR`? As of now the user has to enter airplane mode, open the app and disable airplane mode again.

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/main/SplashActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/main/SplashActivity.kt
@@ -2,9 +2,7 @@ package de.xikolo.controllers.main
 
 import android.content.Intent
 import android.net.Uri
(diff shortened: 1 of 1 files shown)


tobiasrohloff:
I guess this can be checked with the lifecycle state instead of a variable.
https://developer.android.com/topic/libraries/architecture/lifecycle#lc
`activity.lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)`

bengelhaupt:
Nice solution. Any thoughts on the second thing?

PR #155, opened 2019-07-08
Course Items Refactoring to Architecture Components

bengelhaupt:
Fixes #155


app/src/main/java/de/xikolo/viewmodels/section/CourseItemsViewModel.kt
```diff
@@ -0,0 +1,45 @@
...
+    fun markItemVisited(item: Item) {
+        realm.beginTransaction()
+        item.visited = true
+        realm.commitTransaction()
+
+        itemManager.updateItemVisited(item.id)
+    }
+
+    override fun onFirstCreate() {
+        courseDelegate.requestCourse(networkState, false)
```
tobiasrohloff: Why is the course requested again and when? Is it a good idea to pass the network state?

bengelhaupt: The `SectionDownloadHelper` somehow needs the course. Yes, it is not a good idea to pass the NetworkState, I removed it.

PR #156, opened 2019-07-24
RichTextItems and VideoItems refactored to Architecture Components

### LINKED ISSUE #135: Migrate ItemPresenter to Architecture Components
This also affects its subclasses `RichTextPresenter` and `VideoPreviewPresenter`.

bengelhaupt:
Refactors the `RichTextPresenter` and `VideoPreviewPresenter` to Architecture Components.

I had some trouble implementing this. When trying to observe a `LiveData` object of an `Item` (like `RichText`), Realm would complain that the object no longer exists in the database, even though it was requested via `GetItemWithContentJob`, and causes a crash. There was also some legacy code that confused me (and which could be some workaround for this).

Why are we querying all RichTexts and then just copy the first item from Realm? Why not `findFirstAsync()` in the first place?
```
ItemManager>>
fun getRichTextForItem(contentId: String, realm: Realm, listener: RealmChangeListener<RealmResults<RichText>>?): RealmResults<*> {
        if (listener == null) {
            throw IllegalArgumentException("RealmChangeListener should not be null for async queries.")
        }

        // RealmChangeListener for RealmObject doesn't notify for initial copyToRealm
        val richTextPromise = realm
            .where(RichText::class.java)
            .equalTo("id", contentId)
            .findAllAsync()

        richTextPromise.addChangeListener(listener)

        return richTextPromise
    }


RichTextPresenter>>
... richTextPromise = itemManager.getRichTextForItem(item.contentId, realm, (result) -> {
            if (result.size() > 0) {
                richText = realm.copyFromRealm(result.first());
                if (isViewAttached()) {
                    getView().showContent();
                    getView().setupView(item, richText);
                }
            }
        });
```

Fixes #135


tobiasrohloff:
I think this legacy code exists exactly because of the issue you described. I remember I had similar issues and this solved it somehow. Guess we should dig deeper to find the root cause.

PR #164, opened 2019-08-12
Replace RequestJobs with NetworkJobs

### LINKED ISSUE #163: Refactor RequestJobs to NetworkJobs

bengelhaupt:
Replaces the remaining `RequestJobs` with `NetworkJobs`.

I had some problems to find a decent way how to handle the `API_VERSION_DEPRECATED` error with the deprecation date. 
Also, the `SecondScreenManager` is a bit hard to read because of the three nested requests.

Fixes #163


app/src/main/java/de/xikolo/network/jobs/base/NetworkJob.kt
```diff
@@ -33,6 +34,12 @@ abstract class NetworkJob(private val networkState: NetworkStateLiveData, privat
 
     fun error() = networkState.error(userRequest)
 
+    fun deprecated(deprecationDate: Date) = networkState.deprecated(deprecationDate, userRequest)
+
+    fun apiVersionExpired() = networkState.apiVersionExpired(userRequest)
+
+    fun maintenance() = networkState.maintenance(userRequest)
```
tobiasrohloff: Since these methods are only used by `CheckHealthJob` they should be moved there.

app/src/main/java/de/xikolo/network/jobs/base/NetworkStateLiveData.kt
```diff
@@ -17,22 +18,34 @@ class NetworkStateLiveData : LiveData<NetworkState>() {
...
+    fun deprecated(deprecationDate: Date, userRequest: Boolean) {
+        state(NetworkCode.API_VERSION_DEPRECATED, userRequest, deprecationDate)
+    }
+
+    fun apiVersionExpired(userRequest: Boolean) {
+        state(NetworkCode.API_VERSION_EXPIRED, userRequest)
+    }
+
+    fun maintenance(userRequest: Boolean) {
+        state(NetworkCode.MAINTENANCE, userRequest)
```
tobiasrohloff: Same here. How about a sub-class of `NetworkStateLiveData`?

app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.kt
```diff
@@ -165,29 +164,35 @@ class ChannelDetailsFragment : NetworkStateFragment<ChannelViewModel>() {
...
-                        enterCourse(courseId)
-                    }
-                }
-            }
 
-            public override fun onError(code: RequestJobCallback.ErrorCode) {
-                if (view != null) {
-                    hideAnyProgress()
-                    if (code === ErrorCode.NO_NETWORK) {
+        val enrollmentCreationNetworkState = NetworkStateLiveData()
```
tobiasrohloff: Why are those `LiveData` objects not managed by a `ViewModel`?

app/src/main/java/de/xikolo/managers/SecondScreenManager.java
```diff
@@ -63,7 +63,67 @@ public void onWebSocketMessageEvent(WebSocketManager.WebSocketMessageEvent event
                     sectionId = message.payload.get("section_id");
                     courseId = message.payload.get("course_id");
 
-                    courseManager.requestCourseWithSections(courseId, courseCallback(message));
+                    NetworkStateLiveData courseRequestNetworkState = new NetworkStateLiveData();
```
tobiasrohloff: It would be fine for me to remove the whole `SecondScreen` feature. It is currently broken on the backend and the legacy code is awful. What do you think?

app/src/main/java/de/xikolo/extensions/NonNullMediatorLiveData.kt
```diff
@@ -26,3 +26,15 @@ fun <T> LiveData<T>.observe(owner: LifecycleOwner, observer: (t: T) -> Unit): (t
     })
     return observer
 }
+
+// observing stops when the observer function returns true
```
tobiasrohloff: What does it mean when the observer function returns true?

app/src/main/java/de/xikolo/network/jobs/base/NetworkJob.kt
bengelhaupt: That's right.

app/src/main/java/de/xikolo/network/jobs/base/NetworkStateLiveData.kt
bengelhaupt: I tried to subclass `NetworkStateLiveData` including `NetworkState` and `NetworkCode`. The problem with that is, that the `enum class NetworkCode` cannot be inherited from, effectively making subclassing impossible.

app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.kt
bengelhaupt: Each `NetworkStateLiveData` can only be used once for such a request, because when you try to run a new request with an already used `NetworkStateLiveData`, once you call `observe(...)` on it, the `LiveData` object considers a notify with the previous state (which is often `SUCCESS`).

app/src/main/java/de/xikolo/managers/SecondScreenManager.java
bengelhaupt: I can do that. Should I do it in this PR or open a new one?

app/src/main/java/de/xikolo/extensions/NonNullMediatorLiveData.kt
bengelhaupt: The 2nd argument for `observeOnce(owner: LifecycleOwner, observer: (t: T) -> Boolean): (t: T) -> Boolean` is called `observer` and returns a Boolean. When `true` is returned, the observer is removed. 
You usually return `false` for the first observed state `NetworkCode.STARTED` and `true` for `SUCCESS`, `ERROR` etc.

app/src/main/java/de/xikolo/network/jobs/base/NetworkStateLiveData.kt
bengelhaupt: One way of solving this would be to replace the `enum` fields with Int constants but this would reqire changes throughout the entire codebase.

tobiasrohloff: You don't need to subclass `NetworkCode`. Just move it to a separate file. It is fine if it contains values that are not used by the base class `NetworkStateLiveData` and only by a subclass of it.

app/src/main/java/de/xikolo/controllers/channels/ChannelDetailsFragment.kt
tobiasrohloff: Doesn't sound perfect but might be fine for now.

app/src/main/java/de/xikolo/managers/SecondScreenManager.java
tobiasrohloff: Best would be to do that first in a PR and rebase this one after merge.

app/src/main/java/de/xikolo/extensions/NonNullMediatorLiveData.kt
tobiasrohloff: Okay.

PR #166, opened 2019-08-15
Replace EventBus events with LiveData states

### LINKED ISSUE #139: Refactor NetworkChangeReceiver
Our `NetworkChangeReceiver` listens to network changes via `CONNECTIVITY_ACTION` intents. This is deprecated and should be replaced.

### LINKED ISSUE #141: Videos disappear during Section Downloads
When starting a section download including video files, some will be downloaded and some just "disappear". The notification shows the correct amount of file downloads first but then some get lost.

I tested it with section 1 and 2 of this course:  
https://open.hpi.de/courses/mikrocontroller2019/

### LINKED ISSUE #158: Try to migrate connection state changes from EventBus to LiveData object

bengelhaupt:
Replaces `EventBus` events with `LiveData` states.

During implementation, three different kinds of events/states emerged.
- **pure-global** state: like is there a network connection or is the user logged in. Each could easily be replaced by a single global inheritor of a `LiveDataState` object.
- **parameterized** state: for example the current state of a download. With `EventBus`, the event had a parameter and every subscriber received the event regardless of whether it fits into the current context. The `PermissionStateLiveData` and `DownloadStateLiveData` are based on a static dictionary which holds a state for each identifier (e.g. a download url). Via `DownloadStateLiveData.of(URL)` the unique state for that identifier can be queried and observed.
- **signal** events: actually not a state but an event as there is no instance of a state attached to them. Concrete example is the `AllDownloadsCancelled`-Functionality. It is implemented via a `LiveDataEvent` which is basically a `LiveData` with a counter which is incremented on each signaling. 

Other changes:
- `ViewModelMainFragment` was renamed to `MainFragment` and the refresh functionality of login/logout events was moved there from it's subclasses
- The `NetworkChangeReceiver` has been removed and behavior is encapsulated inside the `ConnectivityStateLiveData` class
- `NetworkStateLiveData` now inherits from `LiveDataState`
- `events` package was renamed to `states`

Fixes #158 
Fixes #139 
Maybe fixes #141, because it might have been an issue with stickyness and non-stickyness of events

#
Original Text:
> I did some experiments with `LiveData` objects as a replacement for `EventBus` events.
Instead of subscribing to an Event on the EventBus, the controller observes a `LiveData` state object. This has the advantage of the observing being lifecycle-bound. Events might also not get lost, because active observers get the new state immediately and inactive observers get it when starting to observe.
I found the approach with having a global `state` field in our `App` singleton more robust and probably more performant than writing the state into a preference file and observing changes there. 
I did a first experiment by replacing the `NetworkStateEvent`, which was fired on connectivity changes, with a `ConnectivityEventLiveData` (`ConnectivityStateLiveData` might be a better name though). It works well as far as I can tell.


tobiasrohloff:
That was a lot of work, thanks 🙏

PR #172, opened 2019-08-26
Fix mooc.house play button color

bengelhaupt:


bengelhaupt:
> I don't think that a duplicated layout is needed. You could introduce new feature colors (play button icon and background). These can be overwritten by moochouse then.

Thats right.

PR #173, opened 2019-08-29
Course Root, Recap and Item Deep Link

bengelhaupt:
Introduces a Deep Link for `/courses/...`, `/learn?course_id=...` and `/courses/.../items/...`.

**Courses Deep Link**
Because Android only supports a subset of RegEx functionality in the `pathPattern` attribute in an Intent Filter, now the app accepts any path that starts with `/courses/`. To cope with invalid id's and routes the app has the following behavior:
- Valid id and valid route: `Course Details` tab of course is shown in `CourseActivity`
- Invalid id: An error message is shown, no activity is (visibly) started and an Intent Chooser is displayed to allow the user to re-select an app to start the Intent with (e.g. the browser) (Fig. 1)
- Valid id and valid, non-reachable route (e.g. Learnings) and user logged out: An error message is shown prompting the user to login, the `CourseActivity` is started with the `Course Details` tab as default and a Intent Chooser is presented (Fig. 2)
- Valid id and invalid route and user logged out: An error message is shown prompting the user to login, no activity is started and a Intent Chooser is presented
- Valid id and invalid route and user logged in: A deep link error message is shown, no activity is started and a Intent Chooser is presented 

**Recap**
- Valid id: `Recap` tab of course is shown in `CourseActivity`
- Invalid id: An error message is shown, no activity is (visibly) started and an Intent Chooser is displayed

**Item**
Our `CourseItemsActivity` requires the item's section id and position of the item. This means, the item needs to be fetched prior to that to gather that information. This is done in the `CourseActivity`.
- Valid id and user logged in: The item's tab is selected in the section shown by `CourseItemsActivity`. The `CourseActivity` of the course the item belongs to is not in the back stack.
- Valid id and user logged out: An error message is shown prompting the user to login, the `CourseActivity` of the course is started and a Intent Chooser is presented
- Invalid id: An error message is shown, no activity is (visibly) started and an Intent Chooser is displayed

_Fig. 1_

_Fig. 2_

Fixes #52 

Note: The changes in the testing code fix errors from the dependency update [1e2ba7e](https://github.com/openHPI/xikolo-android/commit/1e2ba7e6f5576b64755db4c2c22ba553c681e763)

PR #174, opened 2019-08-29
Support for external courses

### LINKED ISSUE #169: Display external courses
Display external courses in course lists with a button leading to the external provider (e.g. another app). 
Deep links and intent filters need to be fixed for that to work seamless.

bengelhaupt:
Shows external courses under `All Courses`.

External courses in the list have a banner saying `External` and provide the option to view the course's details in-app or to jump right into the external app/browser.
In the `CourseActivity`, only the `Course Details` tab is shown and the bottom bar allows for visiting the external course. 
Works well with #173 .

  

Fixes #169

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/base/BaseCourseListAdapter.kt
+++ b/app/src/main/java/de/xikolo/controllers/base/BaseCourseListAdapter.kt
@@ -49,7 +49,7 @@ abstract class BaseCourseListAdapter<M>(val fragment: Fragment, private val onCo
         holder.buttonCourseDetails.visibility = View.VISIBLE
         holder.buttonCourseDetails.setOnClickListener { onCourseButtonClickListener?.onDetailButtonClicked(course.id) }
 
-        if(course.external){
+        if (course.external) {
             holder.layout.setOnClickListener { onCourseButtonClickListener?.onDetailButtonClicked(course.id) }
 
             holder.buttonCourseAction.text = App.instance.getString(R.string.btn_external_course)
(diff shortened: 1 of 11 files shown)


app/src/main/java/de/xikolo/controllers/base/BaseCourseListAdapter.kt
```diff
@@ -49,7 +49,13 @@ abstract class BaseCourseListAdapter<M>(val fragment: Fragment, private val onCo
         holder.buttonCourseDetails.visibility = View.VISIBLE
         holder.buttonCourseDetails.setOnClickListener { onCourseButtonClickListener?.onDetailButtonClicked(course.id) }
 
-        if (course.isEnrolled && course.accessible) {
+        if(course.external){
```
tobiasrohloff: Missing whitespaces 🙈

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
```diff
@@ -149,7 +150,12 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
     private fun setupCourse() {
         Crashlytics.setString("course_id", course.id)
 
-        if (!course.isEnrolled) {
+        if(course.external){
```
tobiasrohloff: You don't like whitespaces here? :D

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
```diff
@@ -356,6 +369,11 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
         }
     }
 
+    private fun enterExternalCourse(course: Course){
```
tobiasrohloff: And here..

app/src/main/java/de/xikolo/controllers/main/CourseListFragment.kt
```diff
@@ -217,6 +223,11 @@ class CourseListFragment : ViewModelMainFragment<CourseListViewModel>() {
         startActivity(intent)
     }
 
+    private fun enterExternalCourse(course: Course){
```
tobiasrohloff: and here...

bengelhaupt:
> Can we try to fit the button label in one line? Like "extern öffnen" and "open external". And then the banner could be a bit more descriptive: "External course on open.sap.com" (url host can be taken from external_url?)

I just went with the strings from the web, but that approach will definitely look better.

bengelhaupt:
Ready for review ✓, including whitespaces :)

PR #175, opened 2019-09-15
Bugfixes

### LINKED ISSUE #159: Fix CourseListFragment showing "No Enrollments" error after login although there are enrollments
Also see https://github.com/openHPI/xikolo-android/pull/151#issuecomment-514394458

bengelhaupt:
Fixes three bugs:

- when  opening the `CourseItemsActivity`, the first item is opened but not marked visited
- when in `CourseItemsActivity` and navigating up, the `MainActivity` appears instead of the `CourseActivity` because `onNewActivity` is called which restarts the activity but it should only do this when it is a Deep Link and not an up Intent
- Fixes #159 . The problem was that the courses were synced before the enrollments, so LiveData was triggered when enrollments were not even loaded (only courses are observed, enrollments not) 

(A fourth known bug, a crash when logging out and the `ProfileFragment` is visible, will be fixed in #166)


tobiasrohloff:
Nice, thx :)

PR #176, opened 2019-09-19
NetworkStateHelper Refactorings

### LINKED ISSUE #160: Refactor duplicated network controller helper code into own class/interface
Move common methods such as `hideAnyProgress()`, `showNetworkRequired()` etc. from the respective `NetworkStateFragment` and `BaseViewModelDialog` classes to a helper class or common interface so that code duplication is removed and other classes like the `ViewModelActivity` or `WebViewFragment` can use it. Latter just needs these common methods but does not have a ViewModel associated with it. By now it inherits from a `NetworkStateFragment` and has a NullObject ViewModel which does effectively nothing to workaround this.

bengelhaupt:
The former `NetworkStateFragment` has been split up into the `NetworkStateFragment` and `ViewModelFragment` with latter inheriting from the former. The `NetworkStateFragment` now provides functionality regarding refreshing and the `NetworkStateHelper`. The `ViewModelFragment` builds upon that and adds ViewModel-specific functionality. This allows the `WebViewFragment` to directly inherit from the `NetworkStateFragment`. Thus it does not need a ViewModel anymore and the `NullViewModel` has been removed.
There is now an `interface NetworkStateOwner` which implements shorthand functionality specific to the `NetworkStateHelper`. This helps reducing duplication and separates concerns.

Fixes #160

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/announcement/AnnouncementFragment.kt
+++ b/app/src/main/java/de/xikolo/controllers/announcement/AnnouncementFragment.kt
@@ -10,7 +10,7 @@ import android.widget.TextView
 import butterknife.BindView
 import com.yatatsu.autobundle.AutoBundleField
 import de.xikolo.R
-import de.xikolo.controllers.base.NetworkStateFragment
+import de.xikolo.controllers.base.ViewModelFragment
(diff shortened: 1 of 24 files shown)


tobiasrohloff:
Nice!

PR #177, opened 2019-10-04
Fix for broken video scaling in Exoplayer

bengelhaupt:
Apparently one of the latest dependency updates again broke an Exoplayer feature.
Videos were no longer scaled retaining their aspect ratio but stretched to fit. A flag change fixes this to have the intended behavior of a zoom effect keeping the ratio intact.


tobiasrohloff:
Uh, I will be more careful when upgrading this in the future. But I don't think there was a major update. :/

PR #179, opened 2019-10-04
Resources cleanup

bengelhaupt:
Sorry for that monstrous PR.
I cleaned up and refactored all layout, drawable and color resources.

This includes
- removal of unused files
- consolidation of semantically duplicate resources and integration of `<include>`s with only one usage
- unified naming of layout files
-- _activity\_*_ for layouts of activities
-- _fragment\_*_ for layouts of fragments
-- _dialog\_*_ for layouts of dialogs
-- _item\_*_ for layouts resources of elements appearing in lists
-- _container\_*_ for layouts serving as placeholders or containers
-- _view\_*_ for simple element views
- extraction of all colors that are not `transparent`, `apptheme_main|_second|...` and `text_main|...` from layouts and drawables into color resources
- refactoring of `apptheme_` colors (including wrappers for all of them)
- removal of `?android:attr/` in favor of `?attr/` to always use `appcompat` where available
- removal of custom circular ripples in favor of `?attr/selectableItemBackgroundBorderless`

Also fixes the bug of a black up navigation arrow in the video activity for the `mooc.house` flavor which caused its invisibility because of black on black.


tobiasrohloff:
Awesome :) Only one conflict left.

PR #180, opened 2019-10-08
Resource and Tinting fixes

bengelhaupt:
#179 introduced a bug that made the App crash when clicking on Channels. Apparently only using `?attr/selectableItemBackground` instead of `?android:attr/selectableItemBackground` makes the whole thing crash for some layouts. I have no idea why because it works fine in other places.

Also, the `TintUtil` was somewhat broken because the menu was tinted before the icons were even inflated... I removed the whole thing and achieved the same effect via `app:iconTint` and specific `mooc.house` color resources.


tobiasrohloff:
Nice, didn't know there is something build-in in the meantime.

PR #191, opened 2019-11-24
v3.0 CourseListAdapter crashes

bengelhaupt:
There have been reports on [CourseListAdapter](https://console.firebase.google.com/project/opensap-firebase/crashlytics/app/android:de.xikolo.opensap/issues/a8685041fec7a7ae90e4e13ca055658b) crashes in v3.0. Somehow the data object `DateOverview` ended up being null when fetched via `MetaSectionList.get()`.
I was NOT able to reproduce the issue or find a cause, so I searched what could be the problem. It seems neither to be vendor- nor sdk- nor enrolled-courses-specific.
I found a [StackOverflow post](https://stackoverflow.com/a/30057015) about updating data inside the `RecyclerView.Adapter`. It suggests not setting the list as a new object on each update, but to update the existing list object with the new values. There could be timing, race-condition, lost-update or other similar scenarios with our current approach.
This might be a fix to it but it might as well be not. 

Could you possibly test on your (real) devices with the current production release whether the failure occurs? Else I would suggest rolling out this hotfix to the same users that are currently in the staged rollout phase and see if crashes drop.

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/base/BaseMetaRecyclerViewAdapter.kt
+++ b/app/src/main/java/de/xikolo/controllers/base/BaseMetaRecyclerViewAdapter.kt
@@ -20,10 +20,10 @@ abstract class BaseMetaRecyclerViewAdapter<M, S> : RecyclerView.Adapter<Recycler
         const val ITEM_VIEW_TYPE_ITEM = 2
     }
 
-    protected var contentList: MetaSectionList<String, M, List<S>> = MetaSectionList()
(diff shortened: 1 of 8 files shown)


tobiasrohloff:
Thanks for having an eye on the crash log :) Such an issue never happened on my device but will check it again. Do these changes involve any disadvantage?

bengelhaupt:
> Do these changes involve any disadvantage?

They shouldn't and I saw none while testing.

Issue #190, opened 2019-11-20
Channel Teaser Video


tobiasrohloff:
Certain channels use a teaser video to inform users about the channel. An implementation is needed so that this video is also displayed on the app.

PR #195, opened 2019-12-17
channel teaser video implemented

SanJSp:
This is a first solution for the Channel Teaser Video Issue #190 

The collabsing toolbar is only shown, if there is no Channel Teaser Video. If there is one, the Video will be embedded as meta object in the Meta Section List and the collapsing toolbar is locked. 

~Currently there ist still one major bug:~

~When the ChannelDetailsFragment is opened and one clicks on the channel, the Video is not shown. It is as if there was no test for the stageStream, so the collapsing toolbar is shown. When you reopen the channel the video is shown. Unfortunately the same problem appears when you dont have a teaser video, so a Channel without a stageStream would still lock the toolbar and show the Video Layout. I (and Ben) have not been able to find a solution for that.~

Furthermore I'm looking forward to UI feedback. Here is what it looks like:

You see **two play buttons**, because openSAP embedded a play button on the thumbnail.

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
@@ -8,8 +8,6 @@ import butterknife.BindView
 import com.google.android.material.appbar.AppBarLayout
 import com.google.android.material.appbar.CollapsingToolbarLayout
 import de.xikolo.R
-import de.xikolo.R.id.appbar
-import de.xikolo.R.id.collapsing_toolbar
 import de.xikolo.viewmodels.base.BaseViewModel
 
 abstract class CollapsingToolbarViewModelActivity<T : BaseViewModel> : ViewModelActivity<T>() {
@@ -21,10 +19,10 @@ abstract class CollapsingToolbarViewModelActivity<T : BaseViewModel> : ViewModel
     @BindView(R.id.toolbar_image)
     lateinit var imageView: ImageView
 
-    @BindView(appbar)
+    @BindView(R.id.appbar)
     lateinit var appBarLayout: AppBarLayout
 
-    @BindView(collapsing_toolbar)
+    @BindView(R.id.collapsing_toolbar)
     protected lateinit var collapsingToolbar: CollapsingToolbarLayout
 
     @BindView(R.id.scrim_top)
@@ -47,13 +45,9 @@ abstract class CollapsingToolbarViewModelActivity<T : BaseViewModel> : ViewModel
         lp.height = actionBarHeight + statusBarHeight
 
         collapsingToolbar.isTitleEnabled = false
-
-        if (toolbar != null) {
-            toolbar!!.title = title
-        }
+        toolbar?.title = title
 
         scrimTop.visibility = View.INVISIBLE
-
         scrimBottom.visibility = View.INVISIBLE
     }
 
@@ -70,7 +64,8 @@ abstract class CollapsingToolbarViewModelActivity<T : BaseViewModel> : ViewModel
     private val actionBarHeight: Int
         get() {
             val styledAttributes = theme.obtainStyledAttributes(
-                intArrayOf(android.R.attr.actionBarSize))
+                intArrayOf(android.R.attr.actionBarSize)
+            )
             val result = styledAttributes.getDimension(0, 0f).toInt()
             styledAttributes.recycle()
             return result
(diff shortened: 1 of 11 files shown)


bengelhaupt:
I was able to fix the issue. The problem was that in the `ChannelDetailsActivity`, no network request was made, so the unmanaged database entry had been used which might not yet have a `stageStream` set.
For this, I had to create the `CollapsingToolbarViewModelActivity` which is the `CollapsingToolbarActivity` inheriting from `ViewModelActivity` instead of `BaseActivity`.
There was also a bit of white space around the video, caused by the RecyclerView padding. I introduced a somewhat messy but short workaround to fix this by applying negative margin on the video.

build.gradle
```diff
@@ -13,15 +13,15 @@ buildscript {
...
+        realmSchemaVersion  = 9
     }
     repositories {
         jcenter()
         google()
         maven { url "https://maven.fabric.io/public" }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:3.5.3'
```
tobiasrohloff: No dependency updates in feature branches pls.

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
```diff
@@ -121,6 +121,12 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
 
             oldVersion++;
         }
+
+        if (oldVersion == 8) {
+            schema.get("Channel").addRealmObjectField("stageStream", schema.get("VideoStream"));
```
tobiasrohloff: Please merge with changes on master.

SanJSp: done

app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+        val TAG: String = CollapsingToolbarViewModelActivity::class.java.simpleName
+    }
+
+    @BindView(R.id.toolbar_image)
+    lateinit var imageView: ImageView
+
+    @BindView(appbar)
+    lateinit var appBarLayout: AppBarLayout
+
+    @BindView(collapsing_toolbar)
```
tobiasrohloff: `appbar` and `collapsing_toolbar` don't start with `R.id`, so where do they come from?

app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    protected fun lockCollapsingToolbar(title: String) {
+        appBarLayout.setExpanded(false, false)
+        val lp = appBarLayout.layoutParams as CoordinatorLayout.LayoutParams
+        lp.height = actionBarHeight + statusBarHeight
+
+        collapsingToolbar.isTitleEnabled = false
+
+        if (toolbar != null) {
+            toolbar!!.title = title
```
tobiasrohloff: `toolbar?.title = title`

app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
```diff
@@ -0,0 +1,79 @@
...
+            if (resourceId > 0) {
+                result = application.resources.getDimensionPixelSize(resourceId)
+            }
+            return result
+        }
+
+    private val actionBarHeight: Int
+        get() {
+            val styledAttributes = theme.obtainStyledAttributes(
+                intArrayOf(android.R.attr.actionBarSize))
```
tobiasrohloff: The closing brackets from `obtainStyledAttributes` should be on a new line.

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
```diff
@@ -121,7 +121,7 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
 
             oldVersion++;
         }
-
+
```
tobiasrohloff: We don't need these whitespaces in the diff 😅

app/src/main/java/de/xikolo/controllers/base/CollapsingToolbarViewModelActivity.kt
SanJSp: somehow these landed in the imports. Fixed that.

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
SanJSp: I have no idea why they're in here. In an ancient commit, I accidentally deleted the line, then I fixed it in another commit. As I understand it Github shows that a line has been taken away, but another has been added. Is this my fault or Githubs?

tobiasrohloff: Probably the IDE added the whitespaces, pls remove them and keep the empty line.

SanJSp: Solved

PR #197, opened 2020-01-08
Native LTI and Peer Assessments

### LINKED ISSUE #50: Learning Items Device Suggestions
Some learning items are not optimized for mobile or opened externally. We should better communicate this to learners, by suggesting a proper device advise per item type, where appropriate.

bengelhaupt:
Implements a `LtiExerciseFragment` and `PeerAssessmentFragment`.

The `LtiExerciseFragment` shows a hint `Please note that a mobile device might not be the optimal choice for completing this task.` / `Hinweis: Bitte beachten Sie, dass ein Mobilgerät möglicherweise nicht die optimale Wahl für die Bearbeitung dieser Aufgabe ist.` that the device might not be suited well for completing this task. This message is not shown if the device is a tablet.

  

Fixes #50

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -130,6 +130,11 @@
     <string name="dialog_date_info_title_course">Kursdauer</string>
     <string name="dialog_date_info_title_section">Startdatum</string>
 
+    <string name="dialog_external_content_title">Externer Inhalt</string>
+    <string name="dialog_external_content_message">Diese Aufgabe wird in einem externen Tool geöffnet, welches nicht Teil der openHPI App ist. Dieses ist möglicherweise nicht für Mobilgeräte optimiert, weshalb die Bearbeitung an einem Computer empfohlen wird.\n\nSoll es dennoch geöffnet werden?</string>
+    <string name="dialog_external_content_yes">Öffnen</string>
+    <string name="dialog_external_content_yes_always">Öffnen und nicht erneut nachfragen</string>
+
     <!-- -->
     <string name="title_activity_module">Module</string>
 
@@ -212,10 +217,10 @@
     <string name="course_lti_allowed_attempts">Erlaubte Versuche: %1$s</string>
     <string name="course_lti_allowed_attempts_infinite">Beliebige Versuchsanzahl</string>
     <string name="course_lti_launch">Bearbeitungstool starten</string>
-    <string name="course_lti_device_hint">Diese Aufgabe wird in einem externen Tool geöffnet, welches nicht Teil der openHPI App ist. Dieses ist möglicherweise nicht für Mobilgeräte optimiert, weshalb die Bearbeitung an einem Computer empfohlen wird.</string>
 
     <string name="course_peer_type_solo">Peer Assessment</string>
     <string name="course_peer_type_team">Team peer assessment</string>
+    <string name="course_peer_launch">Peer Assessment öffnen</string>
 
     <string name="document_lang">Sprache: %s</string>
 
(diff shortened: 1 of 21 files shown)


tobiasrohloff:
Please merge with the schema changes on master.

bengelhaupt:
> For LTI, how is this message displayed? As a toast? I think I would prefer it as a dialog to make sure the user won't miss it. I would also prefer a wording like: `This task is opened with an external tool that is not part of the openHPI app. It may not be optimized for mobile devices, so we recommend that you complete it on a computer.`

The current implementation is as a static text below the launch button:




> For peer assessments, I'm not even sure if we should allow to open it with in-app webviews. In most steps, users need to download or upload something and this won't work. So let's show the native info page (when the deadline hasn't passed yet) and say something like `Peer Assessments are not optimized for mobile. Please use the web version on your computer.`

Peer Assessments don't have a launch url like LTIs, thus no button (or am I missing something?: https://dev.xikolo.de/api-docs/#endpoint-peer-assessments).

tobiasrohloff:
Peer Assessments don't need a special URL, you only need the id:
https://open.hpi.de/peer_assessments/399a8fdd-c014-45f1-ab80-8fdf18259fe3

However, what do you think about my approach?

bengelhaupt:
Put the hint in a dialog and add a user preference to hide it in future.
Open LTI link in in-app browser to prevent deep link rerouting.
"Open Peer assessment" button with self-built url.

bengelhaupt:
Done. The Deep Link issue persisted even when using CustomTabs. So I wrote an extension to exclude packages from the chooser.

app/src/main/res/values-de/strings.xml
```diff
@@ -130,6 +130,11 @@
     <string name="dialog_date_info_title_course">Kursdauer</string>
     <string name="dialog_date_info_title_section">Startdatum</string>
 
+    <string name="dialog_external_content_title">Externer Inhalt</string>
+    <string name="dialog_external_content_message">Diese Aufgabe wird in einem externen Tool geöffnet, welches nicht Teil der openHPI App ist. Dieses ist möglicherweise nicht für Mobilgeräte optimiert, weshalb die Bearbeitung an einem Computer empfohlen wird.\n\nSoll es dennoch geöffnet werden?</string>
```
tobiasrohloff: `openHPI` should be replaced with a dynamic flavor name.

app/src/main/res/values/strings.xml
```diff
@@ -132,6 +132,11 @@
     <string name="dialog_date_info_title_course">Course duration</string>
     <string name="dialog_date_info_title_section">Start date</string>
 
+    <string name="dialog_external_content_title">External content</string>
+    <string name="dialog_external_content_message">This task is opened with an external tool that is not part of the openHPI app. It may not be optimized for mobile devices, so we recommend that you complete it on a computer.\n\nDo you want to open it anyway?</string>
```
tobiasrohloff: Same here, `openHPI` should be replaced with a dynamic flavor name.

app/src/main/res/values/strings.xml
```diff
@@ -132,6 +132,11 @@
...
+    <string name="dialog_external_content_yes">Open</string>
+    <string name="dialog_external_content_yes_always">Open and don\'t ask again</string>
```
tobiasrohloff: We use `launch` as the button action for LTIs and `open` for Peer Assessments. Let's make this consistent for the dialog options.

app/src/main/res/values-de/strings.xml
bengelhaupt: Done. Please check the German translation.

PR #199, opened 2020-01-08
Fix for blank screen in CourseItemsActivity upon rotation

bengelhaupt:
I believe this is caused by our messy handling of the fragment adapter (which I do not want to touch again), but this workaround fix works as well.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -154,8 +154,7 @@
         <activity
             android:name=".controllers.section.CourseItemsActivity"
             android:launchMode="singleTop"
-            android:parentActivityName=".controllers.course.CourseActivity"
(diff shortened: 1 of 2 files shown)


bengelhaupt:
As far as I'm concerned it has nothing to do with #205.

I found out it's the line `updateViewPager(listOf())` at the end of `CourseItemsActivity.onCreate()` that causes the bug. It has been introduced in https://github.com/openHPI/xikolo-android/commit/dbeee09752cb55532b7e8812a4aadcdf1ca8f2a3#diff-d40a63043cd4256834516fb97b4933da but I dont quite remember why I made that change. It might be some edge-case.

tobiasrohloff:
Oh, haha, let's see if something else breaks instead 😅

PR #214, opened 2020-02-27
Workaround for bundling VideoStream- and VideoItemPlayerFragment

bengelhaupt:
Argument bundling for the respective Fragments is now handled manually instead of using AutoBundle for the inheritance relation.

@tobiasrohloff Please test whether the build succeeds on your machine.

Will conflict with #195


tobiasrohloff:
Works fine, rebased and fixed some formatting issues.

Issue #210, opened 2020-01-28
Display Time Effort for Course Items


tobiasrohloff:

tobiasrohloff:
iOS looks like this:
![photo_2020-02-25 13 27 41](https://user-images.githubusercontent.com/2922744/75247615-a323c300-57d2-11ea-8bf0-0784cf75702d.jpeg)

PR #216, opened 2020-03-03
Display Time Effort for Course Items implemented

SanJSp:
This is the first solution for displaying the time duration of course items. The solution uses the duration in seconds that is fetched from the API, divides it by 60 and rounds up to the next minute.

Please comment on the layout. I'm not sure if the field timeEffort should be an Integer, because the API returns the duration as an Any. Could this be a problem?

Further ideas may be a display of "isDownloadable" and displaying the achieved points of total points for each item.

Solves #210

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
+++ b/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
@@ -46,7 +46,11 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
 
         holder.textIcon.setText(item.iconRes)
 
-        holder.duration.text = (item.timeEffort / 60 + 1).toString() + "min"
+        if (item.timeEffort > 0) {
+            holder.duration.text = item.formatTimeEffort()
+        } else {
+            holder.duration.visibility = View.GONE
+        }
 
         if (!item.visited) {
             holder.viewUnseenIndicator.visibility = View.VISIBLE
@@ -103,5 +107,4 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
         }
 
     }
-
 }
--- a/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
+++ b/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
@@ -46,6 +46,8 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
 
         holder.textIcon.setText(item.iconRes)
 
+        holder.duration.text = (item.timeEffort / 60 + 1).toString() + "min"
+
         if (!item.visited) {
             holder.viewUnseenIndicator.visibility = View.VISIBLE
         } else {
@@ -92,6 +94,9 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
 
         @BindView(R.id.container)
         lateinit var layout: FrameLayout
+
+        @BindView(R.id.itemDuration)
+        lateinit var duration: TextView
 
         init {
             ButterKnife.bind(this, view)
--- a/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
+++ b/app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
@@ -48,6 +48,7 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
 
         if (item.timeEffort > 0) {
             holder.duration.text = item.formatTimeEffort()
+            holder.duration.visibility = View.VISIBLE
         } else {
             holder.duration.visibility = View.GONE
         }
--- a/app/src/main/java/de/xikolo/models/Item.java
+++ b/app/src/main/java/de/xikolo/models/Item.java
@@ -47,6 +47,8 @@ public class Item extends RealmObject {
 
     public String courseId;
 
+    public Integer timeEffort;
+
     public Section getSection() {
         return SectionDao.Unmanaged.find(sectionId);
     }
@@ -142,6 +144,9 @@ public class Item extends RealmObject {
 
         public HasOne<Course.JsonModel> course;
 
+        @Json(name = "time_effort")
+        public Integer timeEffort;
+
         @Override
         public Item convertToRealmObject() {
             Item item = new Item();
@@ -156,6 +161,7 @@ public class Item extends RealmObject {
             item.proctored = proctored;
             item.visited = visited;
             item.accessible = accessible;
+            item.timeEffort = timeEffort;
 
             if (content != null) {
                 item.contentId = content.get().getId();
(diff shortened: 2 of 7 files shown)


SanJSp:
Latest changes now incorporate the correct display of hour, the display of different languages and multiple line item titles. Additionally, handling for the case that no time duration is given has been implemented.

app/src/main/java/de/xikolo/models/Item.java
```diff
@@ -112,6 +115,26 @@ public int getIconRes() {
...
+
+        if (hours > 0) {
+            if (minutes > 0) {
+                return App.getInstance().getString(R.string.item_duration_hour_minute, hours, minutes);
+            } else {
+                return App.getInstance().getString(R.string.item_duration_hour, hours);
+            }
+        }
+        if (hours == 0) {
+            if (minutes > 0) {
```
tobiasrohloff: This could be a simple `else if` statement without the need to check the hours again.

app/src/main/java/de/xikolo/models/Item.java
```diff
@@ -112,6 +115,26 @@ public int getIconRes() {
...
+                String res = App.getInstance().getString(R.string.item_duration_minute);
+                return String.format(res, minutes);
+            }
+        }
+            return null;
```
tobiasrohloff: Wrong indentation.

app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
```diff
@@ -46,6 +46,12 @@ class ItemListAdapter(private val section: Section, private val listener: OnItem
 
         holder.textIcon.setText(item.iconRes)
 
+        if (item.timeEffort > 0) {
+            holder.duration.text = item.formatTimeEffort()
+        } else {
+            holder.duration.visibility = View.GONE
+        }
```
bengelhaupt: RecyclerView elements might be reused by the Adapter. This can lead to Views being `GONE` initially.
Please add `holder.duration.visibility = View.VISIBLE` in the first branch.

app/src/main/java/de/xikolo/models/Item.java
```diff
@@ -157,6 +183,10 @@ public Item convertToRealmObject() {
             item.visited = visited;
             item.accessible = accessible;
 
+            if (timeEffort != null) {
+                item.timeEffort = timeEffort;
+            }
+
```
bengelhaupt: As the `Item.timeEffort` is `null` by default and nullable, this null-check is redundant.

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
```diff
@@ -147,6 +147,10 @@ public void migrate(DynamicRealm realm, long oldVersion, long newVersion) {
         if (oldVersion == 10) {
             schema.get("Channel").addRealmObjectField("stageStream", schema.get("VideoStream"));
         }
+
+        if (oldVersion == 11) {
+            schema.get("Item").addRealmObjectField("timeEffort", schema.get("time_effort"));
```
bengelhaupt: Should be `int.class` instead of `schema.get("time_effort")` because the field is an Integer.

app/src/main/res/values/strings.xml
```diff
@@ -367,4 +367,9 @@
     <string name="helpdesk_email_error">Please enter a valid email</string>
     <string name="dialog_discard_positive">Discard</string>
     <string name="dialog_discard_inputs">Do you want to discard of your input?</string>
+
+    <string name="item_duration_hour_minute">%1$d hr, %2$d min</string>
```
bengelhaupt: Should there be a comma in the English locale? Because in the German, there isn't.

app/src/main/java/de/xikolo/models/Item.java
tobiasrohloff: The realm field is a primitive `int`, the JSON one a nullable wrapper `Integer`. That's because the API doesn't expose this key for some platforms when this feature is not enabled. On the app side, we're fine with a `0` default value, in which case we don't display the value.

The check is required since you cannot assign a `null` to a primitive `int`.

app/src/main/res/values/strings.xml
tobiasrohloff: True, but why did it work? 🤔

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
tobiasrohloff: True, but why did it work? 🤔

app/src/main/java/de/xikolo/models/Item.java
bengelhaupt: Alright, makes sense. Then we should add the default value.

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
bengelhaupt: RealmMagic©

app/src/main/res/values/strings.xml
tobiasrohloff: It's fine this way. We discussed that in person together with the iOS examples. In German we use abbreviations, in English it's more like a unit.

app/src/main/java/de/xikolo/models/Item.java
tobiasrohloff: It's Java, the default value for `int` is always `0` even without an explicit assignment.

app/src/main/java/de/xikolo/controllers/course/ItemListAdapter.kt
SanJSp: Good to know, thanks! Is fixed

app/src/main/java/de/xikolo/models/Item.java
SanJSp: done, also fixed the string formatting of the other cases

app/src/main/java/de/xikolo/models/migrate/RealmSchemaMigration.java
SanJSp: done

app/src/main/java/de/xikolo/models/Item.java
```diff
@@ -112,6 +115,25 @@ public int getIconRes() {
...
+            } else {
+                return String.format(App.getInstance().getString(R.string.item_duration_hour), hours);
+            }
+        } else if (minutes > 0) {
+            return String.format(App.getInstance().getString(R.string.item_duration_minute), minutes);
+        } else {
+            return null;
+        }
+
+
```
tobiasrohloff: Don't add empty lines at the end of a method.

SanJSp: solved

SanJSp:
Latest commit adds some padding to the items. See screenshot.

Screenshot_1583514813

PR #223, opened 2020-03-23
Hide proctored items

### LINKED ISSUE #217: Disable Proctored Items
If a user booked a qualified certificate (enrollment is proctored) and an item is proctored, we should disable those items and show a warning instead, that this is not possible on mobile.

The web views should check this as well (and I thought they do so), but a user reported that he did proctored stuff on his mobile.

bengelhaupt:
When an Item is proctored, the `ProctoredItemFragment` is inflated into the ViewPager right away. 
Is checking `item.proctored == true` enough?
All this Fragment does is showing the warning

`Nicht verfügbar`
`Dieser Inhalt ist auf Mobilgeräten nicht verfügbar, da ein qualifiziertes Zertifikat gebucht wurde. Bitte nutzen Sie einen Computer, um auf den Inhalt zuzugreifen.`

`Not available`
`This item can't be accessed on mobile because you booked a qualified certificate. Please use a computer to access it.`

I tested the Fragment itself by mocking the check but wasn't able to test the actual check because I don't have a proctored course at hand.

Fixes #217

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/section/CourseItemsActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/section/CourseItemsActivity.kt
@@ -209,16 +209,18 @@ class CourseItemsActivity : ViewModelActivity<CourseItemsViewModel>() {
             var fragment = fragmentManager.findFragmentByTag(name)
             val url = Config.HOST_URL + Config.COURSES + courseId + "/" + Config.ITEMS + item.id
             if (fragment == null) {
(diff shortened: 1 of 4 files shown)


tobiasrohloff:
The general approach looks good to me. However, you need to check the enrollment as well if the proctoring was booked for that course by the user.

bengelhaupt:
I added the enrollment check

tobiasrohloff:
Thanks!

PR #224, opened 2020-03-23
Bugfixes

bengelhaupt:
**Fix for VideoStreamPlayerFragment.kt line 718**
`onDestroy` is called regardless of whether the fragment's view has been created. If this is not the case, the call to the then-null `playerView` will fail.

**Fix for MainActivity.kt line 265, LoginFragment.kt line 229**
The login process consists of two consecutive requests: first the actual login and then fetching the profile. When the user has a slow network connection and leaves the fragment while the first request is running but not the second, the MainActivity would try to operate on a null Profile.
The callbacks of the network requests are correct. When one request errors, the whole process should be undone. But because the observe functions only have the scope of an active fragment, they might not get called when the request errors while the fragment is not active.
When the second is running and the user leaves the fragment, once the observe is invoked, hiding the progress fails because the fragment view has been destroyed.

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/login/LoginFragment.kt
+++ b/app/src/main/java/de/xikolo/controllers/login/LoginFragment.kt
@@ -14,6 +14,7 @@ import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.widget.Toolbar
 import androidx.core.app.NavUtils
+import androidx.lifecycle.Observer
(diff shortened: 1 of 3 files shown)


tobiasrohloff:
Thank you! Any drawbacks here, because you marked it as a draft?

bengelhaupt:
> Thank you! Any drawbacks here, because you marked it as a draft?

No, I just forgot to remove the "Draft status".

PR #225, opened 2020-03-23
Add jump-off points for additional external pages

### LINKED ISSUE #207: Add jump-off points for additional external pages
For each platform, there can be additional external pages - like microlearning or podcasts - which should be integrated into the apps. As of now, we can't receive any of the required data via the API. Therefore, the only option we have is to add jump-off points that link to the respective websites. 

As a first step, we should create a UI mockup of these jump-off points. For the android app, we are likely to include the additional points in the side drawer.
Please create a screenshot of the side drawer showing additional points for `Microlearning` and `Podcasts`.

bengelhaupt:
Just like in picture 3 of #207. Enabled only for openSAP.

Fixes #207


tobiasrohloff:
Since this a openSAP-only feature I would like to keep as much code as possible in the corresponding build folder `opensap`. Do you think it is possible to move it there?

tobiasrohloff:
Okay, Kommando zurück. Podcasts are coming for openHPI as well. So let's build this as a generic "external links" feature for the navigation drawer. Maybe we can keep this out of the `Config` and use brand-specific string array resources instead?

bengelhaupt:
Yeah, can do that. I would implement it somewhat as a per-flavor `R.menu.navigation`. So the url's should be placed in the resources instead of the  `Config`? Will the podcasts url of openHPI be `https://open.hpi.de/podcasts`?

tobiasrohloff:
Right now it is `https://open.hpi.de/pages/podcast`. Yes, I would like to put that into the resources. I'll probably refactor all the legal links in the same way.

bengelhaupt:
I did not go with a per-flavor navigation menu resource because it is a generic feature and the check whether it is supported when setting the url has to be done either way, so I just change the visibility of the items (or the whole group when none are supported).

app/src/main/res/values/strings.xml
```diff
@@ -392,4 +391,8 @@
     <string name="item_duration_hour">%1$d hr</string>
     <string name="item_duration_minute">%1$d min</string>
 
+    <!-- placeholders -->
+    <string name="url_podcasts">empty</string>
```
tobiasrohloff: How about making these real empty strings and if they are `null` hide the fields? If a brand specifies them they are shown. This way we don't need a separate config.

bengelhaupt: Do you mean ""-empty or nonexistent?

tobiasrohloff: Well, we need a key if there is no brand overwrite:
`<string name="url_podcasts" />`
Should work as default or?

bengelhaupt: Because I am getting compile errors when having them nonexistent

bengelhaupt: Alright!

PR #226, opened 2020-03-24
New CI and PR Lint Reporting

tobiasrohloff:
This PR adds a couple of tools and build pipelines. First of all, I added [GitHub Actions](https://github.com/features/actions) as CI which will probably replace Travis at some point. At the moment we keep them in parallel to see how things work out. The new pipeline steps basically trigger [fastlane](https://fastlane.tools/) _lanes_ to build and test the app. We will use fastlane later to generate screenshots and publish releases automatically, but that's future work. New for you is the _lint_ pipeline. It executes [Android Lint](https://developer.android.com/studio/write/lint) and [ktlint](https://github.com/pinterest/ktlint) (wrapped as a [Gradle Plugin](https://github.com/jlleitschuh/ktlint-gradle)). It will report formatting and other violations inline in your future PRs with [Danger](https://danger.systems/ruby/).


tobiasrohloff:
@bengelhaupt @SanJSp let me know if you discover any issues with the new CI steps or if you have any improvements in mind.

Issue #209, opened 2020-01-28
Android App Shortcuts


tobiasrohloff:
https://developer.android.com/guide/topics/ui/shortcuts

Let's discuss possible shortcut actions with @mathebox.

mathebox:
Here is the respective issue for the iOS app: https://github.com/openHPI/xikolo-ios/issues/653

tobiasrohloff:
iOS looks like this:
![photo_2020-02-25 13 31 56](https://user-images.githubusercontent.com/2922744/75247895-44ab1480-57d3-11ea-9eb8-8b6ad2415e18.jpeg)

PR #227, opened 2020-03-25
Implementation of Shortcuts

SanJSp:
Resolves: #209 

Currently, the shortcuts work but the implementation in our codebase so that the parts are reusable got me into trouble. I've added questions to the problematic code snippets in comments.

Maybe someone can jump in and help me. It's some simple observer magic combined with SharedPreferences.


app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+    val recentCourses: LinkedHashSet<Pair<String, String>>?
+        get() {
+            val json = getString(PREF_RECENT_COURSES)
+            val type = object : TypeToken<LinkedHashSet<Pair<String, String>>>() {}.type
+            return Gson().fromJson(json, type)
+        }
+
+    private val shortcutMax = 3
+    val coursesLive: SharedPreferenceLiveData<String> by lazy {
+        // what name does the preference have? What is the sharedPref Parameter supposed to do?
```
SanJSp: I'm supposed to use a name for a preference as a parameter. But is this free for me to decide how this preference is to be named? Other than that I'm not sure if a chose the correct sharedPreferences as a parameter.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+        SharedPreferenceLiveData.SharedPreferenceStringLiveData(PreferenceManager.getDefaultSharedPreferences(App.instance), "recentCourses", "")
+    }
```
SanJSp: I'm supposed to use a name for a preference as a parameter. But is this free for me to decide how this preference is to be named? Other than that I'm not sure if a chose the correct sharedPreferences as a parameter.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+            }
+            courses.size == shortcutMax               -> {
+                courses.remove(courses.last())
+                courses.add(Pair(courseId, title))
+            }
+        }
+
+        putString(PREF_RECENT_COURSES, Gson().toJson(courses))
+        //is this update procedure correct?
+        coursesLive.update(Gson().toJson(courses))
```
SanJSp: For the liveData class I created was no public setValue() available. So I wrote an update method, that calls the setValue(). Is this correct?

app/src/main/java/de/xikolo/App.kt
```diff
@@ -69,6 +77,13 @@ class App : Application() {
         configureRealm()
         configureDefaultSettings()
         configureWebView()
+
+        val recentCoursesObserver = Observer<String> {
+            updateShortcuts()
+        }
+        val recentCourses = RecentCoursesStorage()
+        //who is the owner?
+        recentCourses.coursesLive.observe(this, recentCoursesObserver)
```
SanJSp: There#s an error, that this is not the correct owner. Can somebody help me with that?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: You have to choose that name. Look at the other storages, you can find these strings here: `app/src/main/res/values/keys.xml`.

app/src/main/java/de/xikolo/App.kt
tobiasrohloff: Which interface is expected for the owner? Find that out and look which classes implement that to choose the right owner and right place for that code. Also, this code should be moved to a method.

app/src/main/java/de/xikolo/App.kt
```diff
@@ -102,4 +117,29 @@ class App : Application() {
         CookieManager.getInstance().flush()
     }
 
+    private fun updateShortcuts() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
```
tobiasrohloff: Should e a feature config. Also, the observer code should only be called is this condition is fulfilled.

app/src/main/java/de/xikolo/App.kt
```diff
@@ -102,4 +117,29 @@ class App : Application() {
...
+            val shortcutManager = getSystemService<ShortcutManager>(ShortcutManager::class.java)
+            val recentCourses = RecentCoursesStorage().recentCourses
+            val intentList = mutableListOf<ShortcutInfo>()
+
+            if (recentCourses != null) {
+                for (course in recentCourses) {
```
tobiasrohloff: Isn't that call `null` safe already? (Not sure, pls find out.)

app/src/main/java/de/xikolo/controllers/main/MainActivity.kt
```diff
@@ -33,6 +41,7 @@ import de.xikolo.utils.DeepLinkingUtil
 import de.xikolo.utils.LanalyticsUtil
 import de.xikolo.utils.extensions.checkPlayServicesWithDialog
 import de.xikolo.viewmodels.main.NavigationViewModel
+import java.util.*
```
tobiasrohloff: A lot of added imports but no code here. Always check your diff pls.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+import androidx.preference.PreferenceManager
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import de.xikolo.App
+import de.xikolo.storages.base.BaseStorage
+import de.xikolo.storages.base.SharedPreferenceLiveData
+
+class RecentCoursesStorage : BaseStorage(PREF_RECENT_COURSES, Context.MODE_PRIVATE) {
+
+    val recentCourses: LinkedHashSet<Pair<String, String>>?
```
tobiasrohloff: It would be nice to introduce a `typealias` for `Pair<String, String>` or something similar to have a better readable API. This way you could avoid obscure calls like `course.first` and call something like `course.id` instead.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+        get() {
+            val json = getString(PREF_RECENT_COURSES)
+            val type = object : TypeToken<LinkedHashSet<Pair<String, String>>>() {}.type
+            return Gson().fromJson(json, type)
+        }
+
+    private val shortcutMax = 3
```
tobiasrohloff: This should be a constant.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+    val coursesLive: SharedPreferenceLiveData<String> by lazy {
+        // what name does the preference have? What is the sharedPref Parameter supposed to do?
+        SharedPreferenceLiveData.SharedPreferenceStringLiveData(PreferenceManager.getDefaultSharedPreferences(App.instance), "recentCourses", "")
+    }
+
+    fun addCourse(courseId: String, title: String) {
+        var courses: LinkedHashSet<Pair<String, String>>? = recentCourses
```
tobiasrohloff: You could avoid a lot of checks here if recentCourses wouldn't e nullable.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: You should update the preferences not the live data object.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,55 @@
...
+
+        putString(PREF_RECENT_COURSES, Gson().toJson(courses))
+        //is this update procedure correct?
+        coursesLive.update(Gson().toJson(courses))
+    }
+
+    companion object {
+        private const val PREF_RECENT_COURSES = "pref_recent_courses"
+    }
+}
```
tobiasrohloff: Always keep an empty line at file end.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: The live data object only wraps the preference and offers an observer if the pref value was changed.

SanJSp: So these lines can be deleted. The preferences are updated in the line above.

app/src/main/java/de/xikolo/App.kt
SanJSp: Yes, it is

app/src/main/java/de/xikolo/App.kt
SanJSp: Can you specify feature config? Google was not very helpfull.
Is it something like @RequiresApi?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
SanJSp: done

app/src/main/java/de/xikolo/App.kt
```diff
@@ -63,12 +73,17 @@ class App : Application() {
         instance = this
     }
 
+    @RequiresApi(Build.VERSION_CODES.N_MR1)
```
tobiasrohloff, line 76: Not sure if you wanna tell the compiler the whole `onCreate` method should be stripped before that API version.

app/src/main/java/de/xikolo/App.kt
```diff
@@ -102,4 +117,40 @@ class App : Application() {
...
     }
 
+
+    @RequiresApi(Build.VERSION_CODES.N_MR1)
+    private fun configureShortcuts() {
+        val recentCoursesObserver = Observer<String> {
+            updateShortcuts()
+        }
+
+        val recentCourses = RecentCoursesStorage()
```
tobiasrohloff, line 127: This variable is the storage not the courses, so please call it `recentCoursesStorage`.

app/src/main/java/de/xikolo/config/Feature.kt
```diff
@@ -20,6 +20,10 @@ object Feature {
         Build.VERSION.SDK_INT >= 26 && App.instance.packageManager
             .hasSystemFeature(PackageManager.FEATURE_PICTURE_IN_PICTURE)
 
+    @JvmField
+    val SHORTCUT =
```
tobiasrohloff, line 24: Hmm, I guess I would prefer the plural `SHORTCUTS` here.

app/src/main/java/de/xikolo/config/Feature.kt
```diff
@@ -20,6 +20,10 @@ object Feature {
         Build.VERSION.SDK_INT >= 26 && App.instance.packageManager
             .hasSystemFeature(PackageManager.FEATURE_PICTURE_IN_PICTURE)
 
+    @JvmField
```
tobiasrohloff, line 23: Is that flag actually be used by any Java code? If not, remove the annotation and make it a `const`.

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
```diff
@@ -168,6 +169,9 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
         handleCourseDeepLinkTab(intent)
 
         updateViewPagerTab()
+
+        var recentCourses = RecentCoursesStorage()
```
tobiasrohloff, line 173: Again, that is the storage so please reflect it in the variable name. Also, this is probably a `val`.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,58 @@
...
+class RecentCoursesStorage : BaseStorage(PREF_RECENT_COURSES, Context.MODE_PRIVATE) {
+
+    val recentCourses: LinkedHashSet<RecentCourse>
+        get() {
+            val json = getString(PREF_RECENT_COURSES)
+            val type = object : TypeToken<LinkedHashSet<RecentCourse>>() {}.type
+            if (json == null || type == null) {
+                return LinkedHashSet<RecentCourse>()
+            }
+            println(json)
```
tobiasrohloff, line 24: Remove debug statements.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,58 @@
...
+            return Gson().fromJson(json, type)
+        }
+
+    val coursesLive: SharedPreferenceLiveData<String> by lazy {
+        // what name does the preference have? What is the sharedPref Parameter supposed to do?
```
tobiasrohloff, line 29: Did you choose a proper name? Is the parameter question related to the passing of `""` as a last param?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        putString(PREF_RECENT_COURSES, Gson().toJson(courses))
+        println(Gson().toJson(courses))
+        coursesLive.update(Gson().toJson(courses))
+    }
+
+    companion object {
+        private const val PREF_RECENT_COURSES = "preference_recent_courses"
```
tobiasrohloff, line 56: Is that done in other storages in the same way? Don't we define them in the `res/values/keys.xml`?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,58 @@
...
+                courses.add(Pair(courseId, title))
+            }
+            courses.size == MAX_SHORTCUTS           -> {
+                courses.remove(courses.last())
+                courses.add(Pair(courseId, title))
+            }
+        }
+
+        putString(PREF_RECENT_COURSES, Gson().toJson(courses))
+        println(Gson().toJson(courses))
```
tobiasrohloff, line 51: Debug leftover.

app/src/main/java/de/xikolo/storages/base/SharedPreferenceLiveData.kt
```diff
@@ -0,0 +1,40 @@
+package de.xikolo.storages.base
+
+import android.content.SharedPreferences
+import androidx.lifecycle.LiveData
+
+abstract class SharedPreferenceLiveData<T>(var sharedPrefs: SharedPreferences,
```
tobiasrohloff, line 6: This is copy pasta from SO? Then please leave a comment with a link to the source here.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: Yes it is, nevermind.

app/src/main/res/values/keys.xml
```diff
@@ -27,6 +27,8 @@
     <string name="preference_first_android_4_deprecation_dialog" translatable="false">first_android_4_deprecation_dialog</string>
     <string name="preference_second_android_4_deprecation_dialog" translatable="false">second_android_4_deprecation_dialog</string>
 
+    <string name="preference_recent_courses" translatable="false">recent_courses</string>
```
tobiasrohloff, line 30: Is this used?

SanJSp:
Currently, the shortcuts look like this:


Problems I'm still facing:
- [ ] 
- [ ] 
- [ ]

app/src/main/java/de/xikolo/config/Feature.kt
SanJSp: Can you check if I fixed the const correctly? Not sure if that's what you've meant

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+                courses.add(Pair(courseId, title))
+            }
+            courses.size < MAX_SHORTCUTS -> {
+                courses.add(Pair(courseId, title))
+            }
+            courses.size == MAX_SHORTCUTS -> {
+                courses.remove(courses.last())
+                courses.add(Pair(courseId, title))
+            }
+        }
```
bengelhaupt, line 48: Am I right assuming the newest element is at the last index?
Then following code should do it as well:
```suggestion
        courses = courses.also { it.add(Pair(courseId, title)) }.takeLast(MAX_SHORTCUTS)
```

app/src/main/java/de/xikolo/App.kt
```diff
@@ -102,4 +116,37 @@ class App : Application() {
...
     }
 
+    @RequiresApi(Build.VERSION_CODES.N_MR1)
+    private fun configureShortcuts() {
+        val recentCoursesObserver = Observer<String> {
+            updateShortcuts()
+        }
+
+        val recentCoursesStorage = RecentCoursesStorage()
+        recentCoursesStorage.coursesLive.observeForever(recentCoursesObserver)
```
bengelhaupt, line 126: Observing is never stopped

app/src/main/java/de/xikolo/App.kt
```diff
@@ -102,4 +116,37 @@ class App : Application() {
...
+
+        for (course in recentCourses) {
+            val courseId = course.first
+            val title = course.second
+
+            val intent = CourseActivityAutoBundle.builder().courseId(courseId).build(applicationContext)
+            intent.action = Intent.ACTION_VIEW
+
+            val shortcut = ShortcutInfo.Builder(applicationContext, courseId)
+                .setShortLabel(title)
```
bengelhaupt, line 143: shortLabel: `A concise phrase that describes the shortcut's purpose. When possible, limit the length of the "short description" of a shortcut to 10 characters.`
Aren't our titles a bit longer than that?

SanJSp: Yes, I see that problem too. The IOS screenshot in #209 also uses the full titles. Alternatively, we could use the label "Continue" but then we could only show the latest course. In my opinion it would be better to show the last three courses with a cut off title than only one. What do you think?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
SanJSp: Your assumption is right, but your code allows the same course to be added multiple times, isn't it? In the case that I visit a course 3 times in a row for example. But if it checks first, if the course is already in the list, your code would make a shorter else branch. 

Something like:
```
if (courses.contains(Pair(courseId, title))) {
                courses.remove(Pair(courseId, title))
                courses.add(Pair(courseId, title))
 } else {
                courses = courses.also { it.add(Pair(courseId, title)) }.takeLast(MAX_SHORTCUTS)
}
```

Or is there also a shortcut possible using scope functions? So maybe:
```
courses = courses.also { 
   if (courses.contains(Pair(courseId, title))) {
                courses.remove(Pair(courseId, title))
   } 
   it.add(Pair(courseId, title)) }.takeLast(MAX_SHORTCUTS)
}
```

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.preference.PreferenceManager
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import de.xikolo.App
+import de.xikolo.storages.base.BaseStorage
+import de.xikolo.storages.base.SharedPreferenceLiveData
+
+typealias RecentCourse = Pair<String, String>
+
+const val MAX_SHORTCUTS = 3
```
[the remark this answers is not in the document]
SanJSp: So I've just looked it up, static properties and methods should go into companion objects. Will fix that ^^

app/src/main/java/de/xikolo/App.kt
```diff
@@ -69,6 +79,10 @@ class App : Application() {
         configureRealm()
         configureDefaultSettings()
         configureWebView()
+
+        if (Build.VERSION.SDK_INT >= Feature.SHORTCUTS) {
```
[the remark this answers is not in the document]
SanJSp: Doing this creates a problem for Android Studio. 
"Call requires API level 25(current min is 21)" 
I don't understand this, because we set the check for API level 25 in Feature.SHORTCUTS. Any ideas?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: Try something like:
```kotlin
courses = courses
    .filter { it.first != courseId }
    .also { it.add(Pair(courseId, title)) }
    .reversed()
    .take(MAX_SHORTCUTS)
    .reversed()
```

app/src/main/java/de/xikolo/config/Feature.kt
tobiasrohloff: Yup, fine.

app/src/main/java/de/xikolo/App.kt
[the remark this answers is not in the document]
tobiasrohloff: Ben is right, everything in `Feature` should provide flags, which indicate if features are available or not. If Android Studio / Lint warns you, put this into a separate method (what you should do anyway) and use the [`TargetApi` annotation](https://developer.android.com/reference/android/annotation/TargetApi).

app/src/main/java/de/xikolo/App.kt
tobiasrohloff: [`ShortcutInfo.Builder`](https://developer.android.com/reference/android/content/pm/ShortcutInfo.Builder) also provides  `setLongLabel`. However, since we display up to 3 courses, we cannot omit the title.

tobiasrohloff:
Have you checked all [best practices for shortcuts](https://developer.android.com/guide/topics/ui/shortcuts/best-practices), e.g., the usage of [`reportShortcutUsed`](https://developer.android.com/reference/android/content/pm/ShortcutManager#reportShortcutUsed(java.lang.String))?

SanJSp:
In the recent commit, I've set up the best practices. Apparently the MAX_SHORTCUTS are 4 and I've also adjusted the icons to the design guidelines. Please check, if I've implemented the feedback correctly and that everything is in the right place. Thank you :)

app/src/main/java/de/xikolo/App.kt
```diff
@@ -69,6 +75,10 @@ class App : Application() {
         configureRealm()
         configureDefaultSettings()
         configureWebView()
+        // QUESTION Is this the right place for the dynamicShorcuts call on launch of app?
+        if (Feature.shortcuts()) {
+            configureShortcuts()
+        }
```
SanJSp, line 81: In the best practices, it's said that one should "check the number of objects returned by getDynamicShortcuts() each time you launch your app and re-publish dynamic shortcuts as needed". Is this the right place for that?

app/src/main/java/de/xikolo/config/Feature.kt
```diff
@@ -34,4 +34,8 @@ object Feature {
         }
     }
 
+    @TargetApi(25)
+    fun shortcuts(): Boolean {
+        return Build.VERSION.SDK_INT >= 25
+    }
```
SanJSp, line 40: Is this the correct way to implement this as a function? @tobiasrohloff gave the feedback, that it should be in a function.

tobiasrohloff: Then you misunderstood me. Make this a `const val`. We only provide flags here. Why do you added the `TargetApi` annotation? This flag needs to be called with all API verions.

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
```diff
@@ -168,6 +173,10 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
         handleCourseDeepLinkTab(intent)
 
         updateViewPagerTab()
+
+        if (Feature.shortcuts()) {
```
tobiasrohloff, line 177: Move the feature check to the helper.

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
```diff
@@ -168,6 +173,10 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
...
+            ShortcutHelper().addCourse(applicationContext, course.id, course.title)
```
tobiasrohloff, line 178: Hmm shouldn't this be called in the ViewModels onFirstCreate? @bengelhaupt what do you think?

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+    @RequiresApi(Build.VERSION_CODES.N_MR1)
+    fun addCourse(context: Context, courseId: String, title: String) {
+
+        val recentCoursesStorage = RecentCoursesStorage()
+        recentCoursesStorage.addCourse(courseId, title)
+
+        updateShortcuts(context)
+    }
+
+    @RequiresApi(Build.VERSION_CODES.N_MR1)
```
tobiasrohloff, line 26: Remove the `RequiresApi` annotations. Move all the code that requires a certain API version to an own method and use `TargetApi` there. The other helper methods which call these method(s) do the feature check before. Other parts of the app don't care so let's do that here.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,45 @@
...
+            // filter duplicates
+            .filter { it.first != courseId }
+            .toMutableSet()
+            // add visited course
+            .apply { this.add(Pair(courseId, title)) }
+            // take last MAX_SHORTCUT courses
+            .reversed()
+            .take(MAX_SHORTCUTS)
+            .reversed()
+            .toMutableSet() as LinkedHashSet<RecentCourse>
```
tobiasrohloff, line 37: I would love to see a couple of unit tests for this logic.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,45 @@
...
+            return Gson().fromJson(json, type)
+        }
+
+    fun addCourse(courseId: String, title: String) {
+        var courses: LinkedHashSet<RecentCourse> = recentCourses
+
+        courses = courses
+            // filter duplicates
+            .filter { it.first != courseId }
+            .toMutableSet()
```
tobiasrohloff, line 30: Why is it not mutable in the first place?

app/src/main/java/de/xikolo/config/Feature.kt
SanJSp: Jup, misunderstood that comment: https://github.com/openHPI/xikolo-android/pull/227#discussion_r408111190

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
SanJSp: filter returns a non-mutable List, therefore I had to make it a mutableSet to use the add method in the following line.

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,61 @@
...
+        for (course in recentCourses) {
+            val courseId = course.first
+            val title = course.second
+
+            val intent = CourseActivityAutoBundle.builder().courseId(courseId).build(context)
+            intent.action = Intent.ACTION_VIEW
+
+            val shortcut = ShortcutInfo.Builder(context, courseId)
+                .setShortLabel(title)
+                .setIcon(Icon.createWithResource(context, R.drawable.ic_shortcut_course))
```
tobiasrohloff, line 45: I guess we could use a vector graphic instead of PNGs here like we do it now for other icons as well?

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
SanJSp: @bengelhaupt can you give your feedback on that? :)

app/src/main/java/de/xikolo/App.kt
```diff
@@ -69,6 +71,9 @@ class App : Application() {
         configureRealm()
         configureDefaultSettings()
         configureWebView()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N_MR1) {
```
bengelhaupt, line 74: Shouldn't this be `Feature.SHORTCUTS`?

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
bengelhaupt: It can't be called in `onFirstCreate`, because the call requires a the course title which is only available after the course has been fetched. Furthermore, `setupCourse()` is only called when the course is actually valid (deep link handling), which would not be the case with the viewmodel.

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.os.Build
+import androidx.core.content.ContextCompat
+import androidx.core.content.ContextCompat.getSystemService
+import de.xikolo.R
+import de.xikolo.controllers.course.CourseActivityAutoBundle
+import de.xikolo.storages.RecentCoursesStorage
+
+const val MAX_SHORTCUTS = 4
+
+class ShortcutHelper {
```
bengelhaupt, line 18: Could be an `object` as well. Alternatively you could add the `Context` as a constructor parameter. I would go with object.

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @TargetApi(Build.VERSION_CODES.N_MR1)
+    fun addCourse(context: Context, courseId: String, title: String) {
+
```
bengelhaupt, line 22: redundant newline

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+            intentList.add(shortcut)
+        }
+        shortcutManager?.dynamicShortcuts = intentList
+    }
+
+    @TargetApi(Build.VERSION_CODES.N_MR1)
+    fun configureShortcuts(context: Context) {
+        val shortcutManager =
+            ContextCompat.getSystemService(context, ShortcutManager::class.java)
+        if (shortcutManager?.dynamicShortcuts?.isEmpty()!!) {
```
bengelhaupt, line 58: `.isEmpty() == true` to maintain null safety

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+            ShortcutHelper().updateShortcuts(context)
```
bengelhaupt, line 59: Why construct a new instance here? This is not a static method, is it?

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @TargetApi(Build.VERSION_CODES.N_MR1)
+    fun updateShortcuts(context: Context) {
+        val shortcutManager =
+            getSystemService(context, ShortcutManager::class.java)
+        val recentCourses = RecentCoursesStorage().recentCourses
+        val intentList = mutableListOf<ShortcutInfo>()
+
+        for (course in recentCourses) {
```
bengelhaupt, line 36: Kotlin provides the `.map` function for patterns like this where you need to transform lists of a type to lists of another type.

The usage in this case would be 

```
val intentList = recentCourses.map { course ->
            val courseId = course.first
            val title = course.second

            val intent = CourseActivityAutoBundle.builder().courseId(courseId).build(context)
            intent.action = Intent.ACTION_VIEW

            ShortcutInfo.Builder(context, courseId)
                .setShortLabel(title)
                .setIcon(Icon.createWithResource(context, R.drawable.ic_shortcut_course))
                .setIntent(intent)
                .build()
}
```

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
+package de.xikolo.storages
+
+import android.content.Context
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import de.xikolo.storages.base.BaseStorage
+
+typealias RecentCourse = Pair<String, String>
+
+const val MAX_SHORTCUTS = 4
```
bengelhaupt, line 10: This constant should only be defined once.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+class RecentCoursesStorage : BaseStorage(PREF_RECENT_COURSES, Context.MODE_PRIVATE) {
+
+    companion object {
+        private const val PREF_RECENT_COURSES = "preference_recent_courses"
+
+        fun addCourseToRecentCourses(
+            courseId: String,
+            title: String,
+            courses: LinkedHashSet<RecentCourse>
+        ): LinkedHashSet<RecentCourse> {
```
bengelhaupt, line 21: Does ist specifically have to be a `LinkedHashSet` or would be `Set` sufficient?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+            return courses
+                // filter duplicates
+                .filter { it.first != courseId }
+                .toMutableSet()
+                // add visited course
+                .apply { this.add(Pair(courseId, title)) }
+                // take last MAX_SHORTCUT courses
+                .reversed()
+                .take(MAX_SHORTCUTS)
+                .reversed()
```
bengelhaupt, line 31: The second reverse does not make any sense because the collection is converted to a `Set` afterwards.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+            courses: LinkedHashSet<RecentCourse>
+        ): LinkedHashSet<RecentCourse> {
+            return courses
+                // filter duplicates
+                .filter { it.first != courseId }
+                .toMutableSet()
+                // add visited course
+                .apply { this.add(Pair(courseId, title)) }
+                // take last MAX_SHORTCUT courses
+                .reversed()
```
bengelhaupt, line 29: This reverse is called on a `Set`, is it?

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
+package de.xikolo.controllers.helper
```
bengelhaupt, line 1: It is strictly speaking not a helper for a controller. I think the `utils` package would be more fitting. Mind to rename then.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
bengelhaupt: And please move it inside the class

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+import android.content.pm.ShortcutManager
+import android.graphics.drawable.Icon
+import android.os.Build
+import androidx.core.content.ContextCompat
+import androidx.core.content.ContextCompat.getSystemService
+import de.xikolo.R
+import de.xikolo.controllers.course.CourseActivityAutoBundle
+import de.xikolo.storages.RecentCoursesStorage
+
+const val MAX_SHORTCUTS = 4
```
bengelhaupt, line 16: Please move this constant into the class.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+                .take(MAX_SHORTCUTS)
+                .reversed()
+                .toMutableSet() as LinkedHashSet<RecentCourse>
+        }
+    }
+
+    val recentCourses: LinkedHashSet<RecentCourse>
+        get() {
+            val json = getString(PREF_RECENT_COURSES)
+            val type = object : TypeToken<LinkedHashSet<RecentCourse>>() {}.type
```
bengelhaupt, line 39: Does this work well with Kotlins typealiases? Because Gson is written for Java

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
bengelhaupt: Another question: Should the shortcuts contain all visited courses or only the ones where the user is enrolled?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+typealias RecentCourse = Pair<String, String>
+
+const val MAX_SHORTCUTS = 4
+
+class RecentCoursesStorage : BaseStorage(PREF_RECENT_COURSES, Context.MODE_PRIVATE) {
+
+    companion object {
+        private const val PREF_RECENT_COURSES = "preference_recent_courses"
+
+        fun addCourseToRecentCourses(
```
bengelhaupt, line 17: why this indirection with supplying `courses` as a parameter? Isn't it always `recentCourses`?

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    fun testAddNewCourseToEmpty() {
+        var set = LinkedHashSet<RecentCourse>()
+        val testSet = LinkedHashSet<RecentCourse>()
+        assertTrue(set.size == 0)
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id1", "Course1", set)
+        assertTrue(set.size == 1)
+        testSet.add(Pair("id1", "Course1"))
```
bengelhaupt, line 20: You should use the typealias here

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        val testSet = LinkedHashSet<RecentCourse>()
+        set = RecentCoursesStorage.addCourseToRecentCourses("id1", "Course1", set)
+        testSet.add(Pair("id1", "Course1"))
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id2", "Course2", set)
+        testSet.add(Pair("id2", "Course2"))
+        assertTrue(set == testSet)
+
+        // check order
+        assertTrue(set.last() == Pair("id2", "Course2"))
```
bengelhaupt, line 36: checking order in a set does not make sense

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
bengelhaupt: If you need an ordered collection, use `List` instead

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,79 @@
...
+        var set = LinkedHashSet<RecentCourse>()
+        val fullTestSet = LinkedHashSet<RecentCourse>()
+
+        (1..MAX_SHORTCUTS).forEach {
+            set.add(Pair("id$it", "Course$it"))
+            fullTestSet.add(Pair("id${it + 1}", "Course${it + 1}"))
+        }
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id5", "Course5", set)
+        assertTrue(set.size == 4)
```
bengelhaupt, line 50: `== MAX_SHORTCUTS`

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    fun testAddDuplicateToFull() {
+        var set = LinkedHashSet<RecentCourse>()
+        (1..MAX_SHORTCUTS).forEach { set.add(Pair("id$it", "Course$it")) }
+
+        val fullTestSet = LinkedHashSet<RecentCourse>()
+        fullTestSet.addAll(listOf(Pair("id1", "Course1"), Pair("id3", "Course3"), Pair("id4", "Course4"), Pair("id2", "Course2")))
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id2", "Course2", set)
+        assertTrue(set.size == 4)
```
bengelhaupt, line 75: `== MAX_SHORTCUTS`

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
tobiasrohloff: I think we used a `LinkedHashSet` on purpose, to have unique entries and defined order.

app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
tobiasrohloff: Only the enrolled ones, to "continue learning".

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: We need order and uniqueness of entries, therefore the usage of `LinkedHashSet`.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: But which implementation of `Set`? Ideally, we want to keep a `LinkedHashSet` all the time.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: Probably typealiases are only syntax sugar and dissolved during compile time?

app/src/main/res/drawable/ic_shortcut_course.xml
```diff
@@ -0,0 +1,12 @@
...
+    android:width="48dp"
+    android:height="48dp"
+    android:viewportWidth="48"
+    android:viewportHeight="48">
+  <path
+      android:pathData="M24,46L24,46C11.8,46 2,36.2 2,24v0C2,11.8 11.8,2 24,2h0c12.2,0 22,9.8 22,22v0C46,36.2 36.2,46 24,46z"
+      android:fillColor="#F5F5F5"/>
+  <path
+      android:pathData="M36,17.5c-0.2,0.3 -0.4,0.5 -0.6,0.6C32,21.1 28.5,24 25,26.9c-0.4,0.3 -1.2,0.4 -1.7,0.2c-2.6,-0.9 -5.2,-1.9 -7.8,-2.9c-0.5,-0.2 -0.9,-0.3 -1.7,-0.6c0,0.9 -0.1,1.7 0,2.4c0,0.3 0.4,0.7 0.7,0.9c2.9,1.3 5.8,2.5 8.7,3.8c0.7,0.3 1.2,0 1.7,-0.4c2.8,-2.5 5.7,-5 8.5,-7.5c0.8,-0.7 1.5,-0.7 2.4,0c-0.3,0.4 -0.6,0.6 -0.8,0.9c-3.2,3 -6.5,5.9 -9.7,8.9c-0.6,0.6 -1.1,0.7 -1.9,0.3c-3.5,-1.6 -6.9,-3.1 -10.4,-4.6c-0.8,-0.3 -1.2,-0.8 -1.1,-1.7c0.1,-1.3 -0.1,-2.6 0,-3.8c0,-0.6 0.4,-1.3 0.9,-1.6c3.5,-2.3 7,-4.5 10.5,-6.7c0.4,-0.3 1.1,-0.4 1.5,-0.3c3.6,1 7.1,2.1 10.7,3.2C35.7,17.2 35.8,17.4 36,17.5z"
+      android:fillColor="#353535"/>
```
tobiasrohloff, line 11: I double-checked that and you can use color resources in vector resources on L+. So you can use the primary color here.

app/src/main/res/drawable/ic_shortcut_course.xml
```diff
@@ -0,0 +1,12 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="48dp"
+    android:height="48dp"
+    android:viewportWidth="48"
+    android:viewportHeight="48">
+  <path
+      android:pathData="M24,46L24,46C11.8,46 2,36.2 2,24v0C2,11.8 11.8,2 24,2h0c12.2,0 22,9.8 22,22v0C46,36.2 36.2,46 24,46z"
+      android:fillColor="#F5F5F5"/>
+  <path
+      android:pathData="M36,17.5c-0.2,0.3 -0.4,0.5 -0.6,0.6C32,21.1 28.5,24 25,26.9c-0.4,0.3 -1.2,0.4 -1.7,0.2c-2.6,-0.9 -5.2,-1.9 -7.8,-2.9c-0.5,-0.2 -0.9,-0.3 -1.7,-0.6c0,0.9 -0.1,1.7 0,2.4c0,0.3 0.4,0.7 0.7,0.9c2.9,1.3 5.8,2.5 8.7,3.8c0.7,0.3 1.2,0 1.7,-0.4c2.8,-2.5 5.7,-5 8.5,-7.5c0.8,-0.7 1.5,-0.7 2.4,0c-0.3,0.4 -0.6,0.6 -0.8,0.9c-3.2,3 -6.5,5.9 -9.7,8.9c-0.6,0.6 -1.1,0.7 -1.9,0.3c-3.5,-1.6 -6.9,-3.1 -10.4,-4.6c-0.8,-0.3 -1.2,-0.8 -1.1,-1.7c0.1,-1.3 -0.1,-2.6 0,-3.8c0,-0.6 0.4,-1.3 0.9,-1.6c3.5,-2.3 7,-4.5 10.5,-6.7c0.4,-0.3 1.1,-0.4 1.5,-0.3c3.6,1 7.1,2.1 10.7,3.2C35.7,17.2 35.8,17.4 36,17.5z"
```
tobiasrohloff, line 10: You use the icon for _all courses_ here. Let's use the icon for _my courses_ instead. :)

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
bengelhaupt: It is only defined that `The returned set preserves the element iteration order of the original collection`.

bengelhaupt:
@tobiasrohloff @SanJSp  currently every opened course is added to the recents, no matter whether the user is enrolled or not

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: Sounds fine, or? The element iteration order should be the order of the `LinkedHashSet`. And the tests check the order as well, so it seems to work as expected?

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
SanJSp: No, its applied to a `LinkedHashSet`. But removing the cast to `MutableSet` will break the usage of  `add()` in the apply lambda.

app/src/main/java/de/xikolo/controllers/helper/ShortcutHelper.kt
SanJSp: Cool! Good to know, thank you :)

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
SanJSp: This allows easier testing of the addition of new courses. Otherwise the tests have to be completely rewritten because App.Instance would have to be mocked. Should it nevertheless be changed?

SanJSp:
I've implemented your PR Feedback, the only thing left to do is the BackStack when you go back from a shortcut opened course. I'm kind of stuck there, but I will have a look into the deep link handling. Maybe you can already have a look at it.

I've closed all resolved conversations. Two open conversations left are on the addition of courses [here](https://github.com/openHPI/xikolo-android/pull/227#discussion_r431738739) and [here](https://github.com/openHPI/xikolo-android/pull/227#discussion_r431733824)

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,80 @@
...
+        assertTrue(set == testSet)
+    }
+
+    @Test
+    fun testAddDuplicateToFull() {
+        var set = LinkedHashSet<RecentCourse>()
+        (1..MAX_SHORTCUTS).forEach { set.add(RecentCourse("id$it", "Course$it")) }
+
+        val fullTestSet = LinkedHashSet<RecentCourse>()
+        fullTestSet.addAll(listOf(RecentCourse("id1", "Course1"), RecentCourse("id3", "Course3"), RecentCourse("id4", "Course4"), RecentCourse("id2", "Course2")))
```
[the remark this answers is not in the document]
bengelhaupt: This still exceeds maximum line length

app/src/main/res/drawable/ic_shortcut_my_courses.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="48dp" android:viewportHeight="240"
+    android:viewportWidth="240" android:width="48dp" xmlns:android="http://schemas.android.com/apk/res/android">
+    <path android:fillColor="#F5F5F5" android:fillType="evenOdd" android:pathData="M120.1,119.9m-112,0a112,112 0,1 1,224 0a112,112 0,1 1,-224 0"/>
+    <path android:fillColor="@color/apptheme_primary" android:pathData="M95,61.7c-11.4,0 -24.3,8 -24.3,16.7v94c0,3 2.7,5.7 5.7,5.7h71.1c3,0 5.7,-2.7 5.7,-5.7V84.1c0,-3 -2.7,-6.5 -6.5,-6.5H83.2c-4.2,0 4.2,-8 11,-8h63.5c2.7,0 4.2,1.5 4.2,4.2v94.3c0,4.6 8,1.1 8,-4.2V71.2c0,-4.9 -4.6,-9.5 -10.7,-9.5H95L95,61.7zM111.3,83.7h23.2v63.5l-11.8,-11l-11.4,11.4C111.3,147.6 111.3,83.7 111.3,83.7z"/>
+</vector>
```
tobiasrohloff, line 5: Please check the formatting of our other vector drawables. I'm quite sure that Reformat Code will do that for you.

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        val fullTestSet = LinkedHashSet<RecentCourse>()
+
+        (1..MAX_SHORTCUTS).forEach {
+            set.add(RecentCourse("id$it", "Course$it"))
+            fullTestSet.add(RecentCourse("id${it + 1}", "Course${it + 1}"))
+        }
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id5", "Course5", set)
+        assertTrue(set.size == MAX_SHORTCUTS)
+        assertFalse(set.contains(RecentCourse("id1", "Course1")))
```
tobiasrohloff, line 52: Additionally test that `Course5` is the last element here.

app/src/test/java/de/xikolo/testing/unit/RecentCoursesTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    @Test
+    fun testAddDuplicateToEmpty() {
+        var set = LinkedHashSet<RecentCourse>()
+        val testSet = LinkedHashSet<RecentCourse>()
+
+        set = RecentCoursesStorage.addCourseToRecentCourses("id1", "Course1", set)
+        set = RecentCoursesStorage.addCourseToRecentCourses("id1", "Course1", set)
+        testSet.add(RecentCourse("id1", "Course1"))
+        assertTrue(set == testSet)
```
tobiasrohloff, line 64: Additionally test `assertTrue(set.size == 1)`.

app/src/main/java/de/xikolo/storages/RecentCoursesStorage.kt
tobiasrohloff: I think the tests prove that this is working as expected.

app/src/main/java/de/xikolo/utils/ShortcutUtil.kt
```diff
@@ -0,0 +1,63 @@
...
+            val courseIntent = CourseActivityAutoBundle.builder().courseId(courseId).build(context)
+            courseIntent.action = Intent.ACTION_VIEW
+
+            val mainActivityIntent = Intent(context, MainActivity::class.java)
+            mainActivityIntent.action = Intent.ACTION_VIEW
+
+            ShortcutInfo.Builder(context, courseId)
+                .setShortLabel(title)
+                .setIcon(Icon.createWithResource(context, R.drawable.ic_shortcut_my_courses))
+                .setIntents(arrayOf(mainActivityIntent, courseIntent))
```
tobiasrohloff, line 48: Ah, it is already supported by the API, nice.

PR #228, opened 2020-03-26
Course Filter

### LINKED ISSUE #58: Course Filter
Filter courses for language or whitelisted classifier. Has to be merged with course search somehow.

bengelhaupt:
Adds filters for course search.

The data model from the API is a `Map<String, List<String>>`, I guess. Realm does not allow to save maps and can't query them. This is why I save such a mapping as a list of strings with each string being `<key>:<value1>,<value2>,<value3>,...`. Realm can't query `RealmList<String>` either, so this is where the `RealmStringWrapper`  with a string `value` field comes in. Matching is then done via `realmQuery.like("classifiers.value", "${it.key}:*${it.value},*"`. I think this is the best workaround for our situation.
There`s also no API endpoint for fetching the values of classifiers (except for channels). I just took the values from the websites. This means they need to be updated regularly.
I also fixed some bugs like the search icon disappearing or the normal course list being loaded when the user refreshes while search is open. I also added a message when there are no search results. And when search is opened and the user scrolls, the action bar will no more scroll away with it.

Fixes #58


bengelhaupt:
Thanks for the review, I will work on it.

tobiasrohloff:
That was quick, big thanks!

PR #230, opened 2020-03-31
New Feature Config

tobiasrohloff:
A pretty large refactoring, but most changes are very small and easy. Instead of having different flavor-specific feature configs in `Config` and `FeatureConfig` this PR replaces these with resource-based feature flippers. The defaults can be found in `/app/src/main/res/values/features.xml`, which every flavor can overwrite. We allow features to be directly enabled by providing the corresponding data if required (`string`s and `string-array`s), but also simple flags as `bool`s. If a feature is enabled can be checked with `Feature.enabled("channels")` for example, no matter if this is a `string`, `string-array` or `bool` resource.

This should help to keep the code-base much cleaner in the long-term since more and more flavors are in the pipeline. Flavors can be added/removed without touching `main` at all now. Everything is encapsulated in their flavor-specific source folders. 

Additional stuff:
- Apply whitelisted language config with `resConfigs` for each flavor `app/build.gradle`.
- Get rid of `BuildType`. `Config` still provides the `DEBUG` and `RELEASE` flags.
- Get rid of `BuildFlavor`. No flavor-specific code anymore. Everything is mapped to feature flippers.
- Other features that are not flavor- but hardware-/software-specific are still configured as fields in `Feature`, e.g., `HLS_VIDEO` and `PIP`.
- Provide some convenient `ResourcesExtensions` for easier resource fetching, also from `Context`s.
- Cleanup string resources, e.g., remove unused, add missing translations, etc.


tobiasrohloff:
@bengelhaupt @SanJSp I have to merge this since it blocks other emergency tasks. Feel free to review anyway, I'm happy to implement feedback also at a later point.

bengelhaupt:
LGTM

app/src/openhpi/res/values/strings.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <string name="settings_copyright">&#169; 2012 &#8211; %d Hasso Plattner Institute</string>
+
+    <string name="login_sso" translatable="false">HPI Identity</string>
```
bengelhaupt, line 6: The openHPI app shows `SSO` in german, because `login_sso` is defined as `SSO` in the german `main` flavor, which is used as fallback and not this english variant.

tobiasrohloff: I fixed that for all flavors in https://github.com/openHPI/xikolo-android/commit/bccf4027b686c18fb45dd42ad306f391a59b7a08 and removed the usage of `translatable="false"` for all flavor strings. This should give us lint issues in case we miss that in the future.

PR #232, opened 2020-04-06
Migrate to APIv4

tobiasrohloff:
Deprecated stuff:
- `title` in `videos` (use `title` from `course-items`)
- `slug` in `channels` (use `id`)
- `max_points` in `quizzes` (use `max_points` from `course-items`)
- `first_name` and `last_name` from `user-profile` (use `full_name`)
- `section-progresses` relation in `course-progresses` (use `section_progresses`)
- `course-progress` relation in `section-progresses` (use `course_progress`)

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/main/MainActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/main/MainActivity.kt
@@ -3,7 +3,6 @@ package de.xikolo.controllers.main
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
-import android.util.Log
 import android.view.Menu
 import android.view.MenuItem
 import android.widget.TextView
@@ -14,9 +13,7 @@ import androidx.drawerlayout.widget.DrawerLayout
 import butterknife.BindView
 import com.google.android.material.navigation.NavigationView
 import de.xikolo.App
-import de.xikolo.BuildConfig
 import de.xikolo.R
-import de.xikolo.config.Config
 import de.xikolo.config.Feature
 import de.xikolo.config.GlideApp
 import de.xikolo.controllers.base.BaseFragment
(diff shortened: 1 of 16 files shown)


app/src/main/java/de/xikolo/models/Quiz.java
```diff
@@ -47,7 +45,6 @@ public Quiz convertToRealmObject() {
             quiz.instructions = instructions;
             quiz.timeLimit = timeLimit;
             quiz.allowedAttempts = allowedAttempts;
-            quiz.maxPoints = maxPoints;
```
bengelhaupt, line 50: You forgot to remove lines 35 and 36 in the `Quiz.JsonModel`

```
@Json(name = "max_points")
public int maxPoints;
```

tobiasrohloff: Good catch! Fixed in https://github.com/openHPI/xikolo-android/commit/2b3b5e6420a5e1dbbce44293bb9e5038201a2d89.

bengelhaupt:
This code in `ChannelDao` is causing problems because it still accesses `slug`. The App crashes when opening channel details.
```
override fun find(id: String?): LiveData<Channel> =
        query()
            .beginGroup()
                .equalTo("id", id)
                .or()
                .equalTo("slug", id)
            .endGroup()
            .findFirstAsync()
            .asLiveData()
```

app/src/main/java/de/xikolo/controllers/main/MainActivity.kt
```diff
@@ -284,14 +280,9 @@ class MainActivity : ViewModelActivity<NavigationViewModel>(), NavigationView.On
...
-                headerView.findViewById<TextView>(R.id.textName).text = String.format(
-                    getString(R.string.user_name),
-                    profile?.firstName ?: "",
-                    profile?.lastName ?: ""
-                )
-
-                headerView.findViewById<TextView>(R.id.textEmail).text = profile?.email ?: ""
+                headerView.findViewById<TextView>(R.id.textName).text = user.name
+
+                headerView.findViewById<TextView>(R.id.textEmail).text = user.profile.email
```
bengelhaupt, line 285: `user.profile` can be null in some cases. I added the null check once as a fix. Is there a particular reason you removed it?

PR #240, opened 2020-04-23
Rework Assets

tobiasrohloff:
Okay, I really overdid it this time according to the changed files count. 🙈 

Basically, this PR cleans-up and reworks all resources to minimize the required assets for new brands and to vectorize as much as possible.

Colors:
- The primary and secondary colors are set in the flavor config and their light and dark variants are calculated and generated automatically by build scripts.
- Rename some colors.

Graphics:
- Rework launcher icons. OpenWHO's adaptive icon is now vectorized as well.
- Rework login banners, they are vectorized.
- Rework splash screen banners. Unfortunately, vector graphics don't work nicely here. This is the only thing I'm not happy about since we have to keep them as density-dependent bitmaps and generate them for new brands. 🙁
- Rework profile screen. Remove the "stock photo" header image. Fix the error fallback image. Add button to edit profile in a web view.
- Rework all other in-app icons. They are all vectorized and unified. I also changed some navigation drawer and video player icons.

Themes:
- Migrate from `Theme.AppCompat` to `Theme.MaterialComponents`.
- General clean-up.
- Inherit dialog theme automatically.


tobiasrohloff:
Thank you! :)

I'll check what causes the blurry rendering there.

tobiasrohloff:
> Everything seems good except for all flavors the login banner looks somewhat blurry.

That should be fixed.

PR #243, opened 2020-05-07
Fix nullable dates

tobiasrohloff:


bengelhaupt:
Does this fix the crash encountered in openHPI when clicking on the "Design Thinking 4.0" course? This one has `null` start- and endDates.

tobiasrohloff:
> Does this fix the crash encountered in openHPI when clicking on the "Design Thinking 4.0" course? This one has `null` start- and endDates.

Guess so. All the following errors:
```
Fatal Exception: java.lang.IllegalStateException
course.startDate must not be null
de.xikolo.controllers.course.DescriptionFragment.showDescription (DescriptionFragment.kt:127)
```

PR #244, opened 2020-05-07
Fix hiding of course dates menu icon

### LINKED ISSUE #241: Hide course dates menu item
Only show the course dates menu icon in the action bar of a course, when there are actually course dates.

I thought we already fixed that, so this might be broken again?

bengelhaupt:
The problem was that checking was not done against the number of dates of the specific course but all dates globally. This meant the icon was shown in every course when the user had any upcoming dates.

Fixes #241


tobiasrohloff:
Nice, thanks!

PR #246, opened 2020-05-14
Get native language name from Locale class

tobiasrohloff:


app/src/main/java/de/xikolo/utils/LanguageUtil.kt
```diff
@@ -1,18 +1,18 @@
...
+import java.util.Locale
 
 object LanguageUtil {
-    fun toNativeName(code: String): String {
-        val context = App.Companion.instance
+    @OptIn(ExperimentalStdlibApi::class)
+    fun toNativeName(tag: String): String {
+        val languageTag =
+            if (tag == "cn") { // work-around for bug in the backend
+                "zh"
```
mathebox, line 10: Is it `zh` or `zh-cn`?

tobiasrohloff: `zh`

PR #258, opened 2020-06-11
Add Fullscreen Button to Video Player

### LINKED ISSUE #254: Add Fullscreen Button to Video Player
I think it would be nice to add a fullscreen button to the video player which forces the player to landscape mode and fullscreen similar to the YouTube app. The rotate-to-landscape-for-fullscreen action is not intuitive for all users.

Screenshot_20200605-113729

bengelhaupt:
Fixes #254

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/video/base/BaseVideoPlayerActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/video/base/BaseVideoPlayerActivity.kt
@@ -35,6 +35,8 @@ import de.xikolo.utils.extensions.aspectRatio
 import de.xikolo.utils.extensions.hasDisplayCutouts
 import de.xikolo.utils.extensions.showToast
 import java.util.ArrayList
+import kotlin.math.abs
+import kotlin.math.min
 
 abstract class BaseVideoPlayerActivity : BaseActivity(), VideoStreamPlayerFragment.ControllerInterface {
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/xikolo/controllers/video/base/BaseVideoPlayerActivity.kt
```diff
@@ -153,6 +161,22 @@ abstract class BaseVideoPlayerActivity : BaseActivity(), VideoStreamPlayerFragme
...
+        requestedOrientation = if (isInLandscape()) {
+            isPortrait = true
+            ActivityInfo.SCREEN_ORIENTATION_PORTRAIT
+        } else {
+            isPortrait = false
+            ActivityInfo.SCREEN_ORIENTATION_LANDSCAPE
+        }
+        Handler().postDelayed({
+            requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+        }, 3000) // user has 3 seconds to rotate device, otherwise it will change back
```
tobiasrohloff, line 175: This should force the orientation, no matter what orientation the device is actually in, just like the YouTube app does. This makes it possible to switch to fullscreen even when auto-rotate is off. This state is even maintained there so that when the app goes into the background (device back to portrait) and is opened again, the forced landscape orientation is restored. Have a look at the behavior of this app in detail, I find it really useful.

bengelhaupt: Implemented rotation like in the YouTube app

tobiasrohloff: The experience is much better now! Unfortunately, I also haven't found a cleaner solution to implement this.

One last question: when auto-rotate is on, the user switches to full/small screen with the button and then rotates the device, the layout is flipped two times again. Can we avoid that?

tobiasrohloff: >One last question: when auto-rotate is on, the user switches to full/small screen with the button and then rotates the device, the layout is flipped two times again. Can we avoid that?

@bengelhaupt What is the status here?

bengelhaupt: > > One last question: when auto-rotate is on, the user switches to full/small screen with the button and then rotates the device, the layout is flipped two times again. Can we avoid that?
> 
> @bengelhaupt What is the status here?

I can't reproduce that behavior on my device. Do you mean that after the screen has rotated through the button and then the device is rotated the same way, the screen flips back to normal and then into the rotated state?

tobiasrohloff: Exactly.

bengelhaupt: Okay, I can reproduce it on the emulator. I will have a look onto it. Did it happen on a real device for you?

tobiasrohloff: Yes, on my Pixel 3.

bengelhaupt: I rewrote the custom rotation detection. Can you test if this works on your device?

tobiasrohloff: Alright, now it works sometimes, still very sensitive. Anyway, I'm fine with it and I guess it is not worth it to optimize this further atm. Thank you!

PR #268, opened 2020-08-12
Fix deeplink issues

### LINKED ISSUE #260: Invalid Deep Links for Course Items
It seems that deep links for course items need to be checked more strictly, as our backend detected the following invalid requests:
Screenshot from 2020-08-10 13-17-50
Due to `DeepLinkUtil`'s removal of slashes, we end up with these weird IDs. I'm not sure if we support any nested links like `courses/*/items/*/another_action` or if this needs to be avoided at all. Anyway, please fix this and add tests.

bengelhaupt:
The matching has been concretized and the parsing has been fixed. Tests have been added accordingly.

Note: When the function returns `null`, it means the link cannot be handled by the app.

Fixes #260

### DIFF
--- a/app/src/androidTest/java/de/xikolo/testing/instrumented/unit/DeepLinkingUtilTest.kt
+++ b/app/src/androidTest/java/de/xikolo/testing/instrumented/unit/DeepLinkingUtilTest.kt
@@ -15,28 +15,90 @@ class DeepLinkingUtilTest : BaseTest() {
     fun testCourseIdentification() {
         assertEquals(
             "123456",
-            DeepLinkingUtil.getCourseIdentifier(Uri.parse("https://open.hpi.de/courses/123456"))
+            DeepLinkingUtil.getCourseIdentifier(
+                Uri.parse("https://open.hpi.de/courses/123456")
+            )
         )
 
         assertEquals(
             "123456",
-            DeepLinkingUtil.getCourseIdentifier(Uri.parse("https://open.hpi.de/courses/123456/resume"))
+            DeepLinkingUtil.getCourseIdentifier(
+                Uri.parse("https://open.hpi.de/courses/123456/")
+            )
         )
 
         assertEquals(
             "123456",
-            DeepLinkingUtil.getCourseIdentifier(Uri.parse("https://open.hpi.de/courses/123456/items/654321"))
+            DeepLinkingUtil.getCourseIdentifier(
+                Uri.parse("https://open.hpi.de/courses/123456/resume")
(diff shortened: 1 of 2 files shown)


app/src/androidTest/java/de/xikolo/testing/instrumented/unit/DeepLinkingUtilTest.kt
```diff
@@ -15,28 +15,90 @@ class DeepLinkingUtilTest : BaseTest() {
...
+            "123456",
+            DeepLinkingUtil.getCourseIdentifier(
+                Uri.parse("https://open.hpi.de/courses/123456/resume")
+            )
+        )
+
+        assertEquals(
+            "123456",
+            DeepLinkingUtil.getCourseIdentifier(
+                Uri.parse("https://open.hpi.de/courses/123456/resume/")
```
tobiasrohloff, line 40: You test the trailing `/` here?

app/src/main/java/de/xikolo/utils/DeepLinkingUtil.kt
```diff
@@ -23,33 +23,64 @@ object DeepLinkingUtil {
...
+     * Course resume path /courses/{course_id}/resume
+     * Course pinboard path /courses/{course_id}/pinboard
+     * Course progress path /courses/{course_id}/progress
+     * Course announcements path /courses/{course_id}/announcements
+     * Course items path /courses/{course_id}/items/.*
+     *
+     * Ignores:
+     * Invalid action /courses/{course_id}/something_invalid
+     * Invalid action path /courses/{course_id}/resume/something
+     * Item sub paths /courses/{course_id}/items/{item_id]/something
```
tobiasrohloff, line 38: `{item_id]` wrong closing bracket :D

app/src/androidTest/java/de/xikolo/testing/instrumented/unit/DeepLinkingUtilTest.kt
bengelhaupt: Yes, is there a problem with this?

tobiasrohloff: No, it just took me a minute to find the difference to the test on top. Maybe a comment about that would be helpful.

bengelhaupt: Done

PR #272, opened 2020-08-18
Replace Bypass with Markwon as used markdown library

### LINKED ISSUE #264: BypassGlideImageGetter: IllegalArgumentException
https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/05dd68714825e744ff96ea5eaadf297f

### LINKED ISSUE #266: TextViewExtensions: IllegalArgumentException
https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/fa2699ff5afb401e3c48c824c1cd7e9e

### LINKED ISSUE #267: TextViewExtensions: RejectedExecutionException
https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/08aa39b64f49cda7514006380757f481

bengelhaupt:
Bypass as our markdown library was not well maintained and required some glue code to work with Glide.
[Markwon](https://github.com/noties/Markwon) is well maintained and based on [commonmark](https://github.com/atlassian/commonmark-java) by Atlassian. It supports Glide and custom styling.

There are only subtle changes in styling such as bullet point size. Markwon would render an underline under headings, which I removed (to match the web). Is there a xikolo markdown reference to compare our rendered content to the web rendered?

Probably 
Fixes #264 
Fixes #266 
Fixes #267 

(I also added a larger padding to the bottom of the course description to better support devices with rounded edges)

### DIFF
--- a/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
+++ b/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
@@ -41,7 +41,6 @@ fun <T : TextView> T.setMarkdownText(markdown: String?) {
             return url
         }
 
-        val glide = Glide.with(context)
         val formatter = Markwon.builder(context)
             .usePlugin(MarkwonInlineParserPlugin.create())
             .usePlugin(JLatexMathPlugin.create(textSize))
@@ -51,6 +50,8 @@ fun <T : TextView> T.setMarkdownText(markdown: String?) {
             .usePlugin(HtmlPlugin.create())
             .usePlugin(
                 GlideImagesPlugin.create(object : GlideImagesPlugin.GlideStore {
+                    private val glide = Glide.with(context)
+
                     override fun cancel(target: Target<*>) {
                         glide.clear(target)
                     }
--- a/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
+++ b/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
@@ -12,6 +12,7 @@ import de.xikolo.config.Config
 import io.noties.markwon.AbstractMarkwonPlugin
 import io.noties.markwon.Markwon
 import io.noties.markwon.core.MarkwonTheme
+import io.noties.markwon.ext.tables.TablePlugin
 import io.noties.markwon.image.AsyncDrawable
 import io.noties.markwon.image.glide.GlideImagesPlugin
 import java.net.MalformedURLException
@@ -35,19 +36,23 @@ fun <T : TextView> T.setMarkdownText(markdown: String?) {
             return url
         }
 
+        val glide = Glide.with(context)
         val formatter = Markwon.builder(context)
             .usePlugin(
                 GlideImagesPlugin.create(object : GlideImagesPlugin.GlideStore {
                     override fun cancel(target: Target<*>) {
-                        Glide.with(context).clear(target)
+                        glide.clear(target)
                     }
 
                     override fun load(drawable: AsyncDrawable): RequestBuilder<Drawable> {
-                        return Glide.with(context).load(
+                        return glide.load(
                             getAbsoluteUrl(drawable.destination)
                         )
                     }
                 })
+            )
+            .usePlugin(
+                TablePlugin.create(context)
             )
             .usePlugin(
                 object : AbstractMarkwonPlugin() {
--- a/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
+++ b/app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
@@ -12,9 +12,14 @@ import de.xikolo.config.Config
 import io.noties.markwon.AbstractMarkwonPlugin
 import io.noties.markwon.Markwon
 import io.noties.markwon.core.MarkwonTheme
+import io.noties.markwon.ext.latex.JLatexMathPlugin
+import io.noties.markwon.ext.strikethrough.StrikethroughPlugin
 import io.noties.markwon.ext.tables.TablePlugin
+import io.noties.markwon.ext.tasklist.TaskListPlugin
+import io.noties.markwon.html.HtmlPlugin
 import io.noties.markwon.image.AsyncDrawable
 import io.noties.markwon.image.glide.GlideImagesPlugin
+import io.noties.markwon.inlineparser.MarkwonInlineParserPlugin
 import java.net.MalformedURLException
 import java.net.URI
 import java.net.URISyntaxException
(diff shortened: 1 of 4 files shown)


tobiasrohloff:
Cool stuff. We don't really have a markdown reference, but this [item](https://mooc.house/courses/kic-howto-course2020/items/30EOw6FL8tvJUIekCn3p5z) has a lot of examples. But it is not bad if everything does not look the same.

PS: Please assign tickets you're working on to yourself, so nobody else will start with these as well.

tobiasrohloff:
Looks nice! I checked the available plugins and I would probably include more of them to provide the best compatibility:
```
implementation "io.noties.markwon:core:$markwon_version"
implementation "io.noties.markwon:ext-latex:$markwon_version"
implementation "io.noties.markwon:ext-strikethrough:$markwon_version"
implementation "io.noties.markwon:ext-tables:$markwon_version"
implementation "io.noties.markwon:ext-tasklist:$markwon_version"
implementation "io.noties.markwon:html:$markwon_version"
implementation "io.noties.markwon:image:$markwon_version"
implementation "io.noties.markwon:image-glide:$markwon_version"
```
What do you think?

app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
```diff
@@ -1,162 +1,90 @@
...
+                GlideImagesPlugin.create(object : GlideImagesPlugin.GlideStore {
+                    override fun cancel(target: Target<*>) {
+                        glide.clear(target)
                     }
 
-                    val drawable = BitmapDrawable(textView.resources, bitmap)
-                    val aspectRatio = 1.0 * drawable.intrinsicWidth / drawable.intrinsicHeight
-
-                    // real image width in pixel scaled based on screen density
-                    val scaledWidth = Math.round(drawable.intrinsicWidth * Resources.getSystem().displayMetrics.density)
```
tobiasrohloff, line 127: I remember that I implemented that to keep the image proportions similar to the web. Is that handled well by the lib out of the box?

E.g., images with a large width like in this [course description](https://open.hpi.de/courses/prototype2019) or [this](https://open.hpi.de/courses/international-teams2019).

bengelhaupt:
> Looks nice! I checked the available plugins and I would probably include more of them to provide the best compatibility:
> 
> ```
> implementation "io.noties.markwon:core:$markwon_version"
> implementation "io.noties.markwon:ext-latex:$markwon_version"
> implementation "io.noties.markwon:ext-strikethrough:$markwon_version"
> implementation "io.noties.markwon:ext-tables:$markwon_version"
> implementation "io.noties.markwon:ext-tasklist:$markwon_version"
> implementation "io.noties.markwon:html:$markwon_version"
> implementation "io.noties.markwon:image:$markwon_version"
> implementation "io.noties.markwon:image-glide:$markwon_version"
> ```
> 
> What do you think?

Sounds good. I added all of them except `image`, because we already have `image-glide`.

app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
bengelhaupt: I observed the behavior, that it does not expand past the container width if it is larger but keeps its original size (is not stretched) when it is smaller than container width.

tobiasrohloff: Looks good. 👍

app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
```diff
@@ -1,162 +1,98 @@
...
-            override fun onPostExecute(bitmap: Bitmap) {
-                val textView = textViewWeakReference.get() ?: return
-                try {
-                    if (maxWidth == -1) {
-                        val horizontalPadding = textView.paddingLeft + textView.paddingRight
-                        maxWidth = textView.measuredWidth - horizontalPadding
-                        if (maxWidth == 0) {
-                            maxWidth = Integer.MAX_VALUE
-                        }
+        val glide = Glide.with(context)
```
tobiasrohloff, line 44: This way it uses the same configured `glide` that we use at other places, right?

However, I would move this into the plugin methods like in the [example](https://noties.io/Markwon/docs/v4/image-glide/).

tobiasrohloff (changes requested):

app/src/main/java/de/xikolo/utils/extensions/TextViewExtensions.kt
bengelhaupt: I had it that way before and it crashed sometimes. If you look at the default implementation of `GlideImagesPlugin.create(...)` [here](https://github.com/noties/Markwon/blob/master/markwon-image-glide/src/main/java/io/noties/markwon/image/glide/GlideImagesPlugin.java), you see this hint:

```
@NonNull
    public static GlideImagesPlugin create(@NonNull final Context context) {
        // @since 4.5.0 cache RequestManager
        //  sometimes `cancel` would be called after activity is destroyed,
        //  so `Glide.with(context)` will throw an exception
        return create(Glide.with(context));
    }
```

tobiasrohloff: But wouldn't it be cleaner to check if the context is still valid beforehand? 🤔  (If this is possible.)

bengelhaupt: we could place everything inside a try-catch block but I would go with the default implementation

bengelhaupt: It could also be placed inside the interface implementation object

bengelhaupt: > It could also be placed inside the interface implementation object

I made a commit for this. I think it achieves your requirement

tobiasrohloff: Silently swallowing exceptions is also not so nice. I would prefer to avoid this by checking the state of context before.

bengelhaupt: But checking the context validity would require to look into what Glide specifies as "valid", which might change between versions

tobiasrohloff: Who throws the exception based on which condition? As far as I understand this happens when retrieving the Glide Singleton, right? And not during the `clear` operation? Because the latter would still have this problem now.

bengelhaupt: The problem is that `clear` is sometimes called when the activity has already been destroyed and `Glide.with(context)` fails.
When placing the `Glide.with(context)` inside the `clear` method, the app will sometimes crash. But if it is placed outside, e.g. in the object's scope, the `glide` variable is initialized when the object is initialized (and the activity is running), leaving a working Glide instance in `glide` to even use in `clear` when `Glide.with(context)` would fail, because it has been cached.

Is it that what you mean?

tobiasrohloff: Yup, I guess I'm fine with it. Having the glide retrieval inside the anonymous plugin object is much cleaner.

SanJSp (approved):
I would not have found that solution only from the firebase error stacks :D good to know about markwon and from my side everything is ok 👍

tobiasrohloff (approved):

PR #273, opened 2020-09-07
Locale LanguageFilter translation implemented

### LINKED ISSUE #247: Improve Course Language Filter
Let's add the language name in the [device's primary locale](https://stackoverflow.com/a/39639103/4870048) as a subtitle to the course language filter as iOS is doing it. We're currently only displaying the language name in its native locale.

Also, please cleanup `CourseFilterView` and `CourseFilterClassifierView`, e.g., the workaround for wrong Chinese API codes should be handled by a new method in `LanguageUtil`. Probably, also a custom layout is required for this filter.

Screen Shot 2020-05-14 at 17 09 06

SanJSp:
Fixes #247 

Changes added to support a translation of course languages to the locale language. The format is nativeLanguage (localLanguage) f.e. Espagnol (Spanisch). Additionally the `LanguageUtil` was cleaned up.

### DIFF
--- a/app/src/main/java/de/xikolo/utils/LanguageUtil.kt
+++ b/app/src/main/java/de/xikolo/utils/LanguageUtil.kt
@@ -1,18 +1,34 @@
 package de.xikolo.utils
 
+import android.content.Context
 import java.util.Locale
 
 object LanguageUtil {
+    var deviceLanguage: String = Locale.getDefault().language.takeIf { it != "zh" } ?: "cn"
+
     @OptIn(ExperimentalStdlibApi::class)
     fun toNativeName(tag: String): String {
-        val languageTag =
-            if (tag == "cn") { // work-around for bug in the backend
-                "zh"
-            } else {
-                tag
-            }
+        val languageTag = correctLanguageTag(tag)
 
         val locale = Locale.forLanguageTag(languageTag)
         return locale.getDisplayLanguage(locale).capitalize(locale)
+    }
+
+    @OptIn(ExperimentalStdlibApi::class)
+    fun toLocaleName(context: Context, tag: String): String {
+        val languageTag = correctLanguageTag(tag)
+        val locale = Locale.forLanguageTag(languageTag)
+        val deviceLocale = Locale.getDefault()
+
+        return locale.getDisplayLanguage(deviceLocale).capitalize(locale)
+    }
+
+    private fun correctLanguageTag(tag: String): String {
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/xikolo/utils/LanguageUtil.kt
```diff
@@ -1,18 +1,40 @@
...
     }
+
+    @OptIn(ExperimentalStdlibApi::class)
+    fun toLocaleName(context: Context, tag: String): String {
+        val languageTag = correctLanguageTag(tag)
+        val locale = Locale.forLanguageTag(languageTag)
+
+        val deviceLocale =
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                context.resources.configuration.locales.get(0)
```
bengelhaupt, line 25: You are using the `context` here to fetch the device language, yet in `deviceLanguage` you are using `Locale.getDefault()`. I think this should be consistent.

bengelhaupt (changes requested):
Simple, yet effective I think. Good work!

app/src/main/java/de/xikolo/utils/LanguageUtil.kt
SanJSp: Found a solution to use even less code :) maybe you can have a look

app/src/main/java/de/xikolo/utils/LanguageUtil.kt
```diff
@@ -1,18 +1,34 @@
...
 
         val locale = Locale.forLanguageTag(languageTag)
         return locale.getDisplayLanguage(locale).capitalize(locale)
     }
+
+    @OptIn(ExperimentalStdlibApi::class)
+    fun toLocaleName(context: Context, tag: String): String {
+        val languageTag = correctLanguageTag(tag)
+        val locale = Locale.forLanguageTag(languageTag)
+        val deviceLocale = Locale(Locale.getDefault().language)
```
bengelhaupt, line 21: Is this necessary, wouldn't `Locale.getDefault()` suffice?

bengelhaupt (changes requested):

app/src/main/java/de/xikolo/utils/LanguageUtil.kt
SanJSp: Yes, it suffices :) good catch!

tobiasrohloff (approved):

PR #277, opened 2020-09-15
Update ExoPlayer to Version 2.12.0

### LINKED ISSUE #275: Update ExoPlayer to Version 2.12.0
The update includes breaking changes and I've stumbled upon a lot of deprecations. Please take care. 💐

bengelhaupt:
Fixes #275

### DIFF
--- a/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
+++ b/app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
@@ -6,8 +6,19 @@ import android.media.MediaMetadataRetriever
 import android.net.Uri
 import android.util.AttributeSet
 import androidx.annotation.IntRange
-import com.google.android.exoplayer2.*
-import com.google.android.exoplayer2.source.*
+import com.google.android.exoplayer2.C
+import com.google.android.exoplayer2.ExoPlaybackException
+import com.google.android.exoplayer2.MediaItem
+import com.google.android.exoplayer2.PlaybackParameters
+import com.google.android.exoplayer2.Player
+import com.google.android.exoplayer2.Renderer
+import com.google.android.exoplayer2.SimpleExoPlayer
+import com.google.android.exoplayer2.Timeline
+import com.google.android.exoplayer2.source.MediaSource
+import com.google.android.exoplayer2.source.MergingMediaSource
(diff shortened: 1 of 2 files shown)


app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
```diff
@@ -176,9 +202,16 @@ open class ExoPlayerVideoView : PlayerView {
         this.uri = uri
         videoMediaSource =
             if (isHls) {
-                HlsMediaSource.Factory(dataSourceFactory).createMediaSource(uri)
+                HlsMediaSource.Factory(dataSourceFactory)
+                    .setAllowChunklessPreparation(true)
```
tobiasrohloff, line 206: That's an `HLS` feature?

app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
```diff
@@ -166,8 +193,7 @@ open class ExoPlayerVideoView : PlayerView {
         exoplayer.setPlaybackParameters(
             PlaybackParameters(
                 speed,
-                exoplayer.playbackParameters.pitch,
-                exoplayer.playbackParameters.skipSilence
```
tobiasrohloff, line 170: The `skipSilence` is not required anymore? I know it is not part of the constructor anymore, but does it need to be set somehow else?

tobiasrohloff (approved):
Thank you! I have only 2 questions for understanding some changes.

app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
bengelhaupt: Yes, it is an optimization allowing HLS videos to load faster by not downloading a chunk first to check codecs etc., because this information is already in the master playlist. Does not matter in the end, because HLS is yet only enabled in debug mode.

app/src/main/java/de/xikolo/views/ExoPlayerVideoView.kt
bengelhaupt: SkipSilence was never required to be set, that's why the new PlaybackParams with the updated speed includes the old value, just like for the pitch.

PR #280, opened 2020-09-21
Displaying of video description in preview implemented

### LINKED ISSUE #252: Display Video Subtitles and Description also in Preview
Screenshot_20200604-093606

SanJSp:
Fixes #252 

This is a first implementation for the display of subtitles and description in a video preview. I added the list of available subtitles, that are translated to the devices locale, a header and some expandable functionality to show the description of the video. When there are less than 130 characters in the description, the Show More button is not visible. I could need some help regarding following issues:

- [ ] how to get rid of magic numbers (lines 240 following in `VideoPreviewFragment.kt`)
- [ ] feedback regarding naming of ressources and variables
- [ ] feedback on the abstraction level in the code, that I tried to keep readable in `VideoPreviewFragment.kt` and `VideoDescriptionDelegate.kt`

Current look of the solution. First image without a description, second with a short description and 3rd and 4th shows how the description expands.

Looking forward to your feedback! Maybe @mathebox also wants to take a look at the solution, dont want to be beaten up again 😅


app/src/main/res/layout/fragment_video_preview.xml
```diff
@@ -25,18 +25,63 @@
...
+                android:textSize="14sp" />
+
+            <TextView
+                android:id="@+id/textPreviewDescription"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_below="@+id/headerTextPreviewDescription"
+                android:layout_alignParentStart="true"
+                android:ellipsize="marquee"
+                android:foreground="@drawable/gradient_light_to_transparent_from_bottom"
```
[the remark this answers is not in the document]
SanJSp: What should I do about this? This means that everybody with an API level below 23 will not see a fade in the description, if it's collapsed. Is that that bad? I dont think so.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -188,4 +224,52 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+
+        if (textDescription.text.length < 115) {
+            showDescriptionButton.visibility = View.GONE
+            if (Feature.FOREGROUNDS) {
+                textDescription.foreground = null
+            }
+        } else if (Feature.FOREGROUNDS) {
+            textDescription.foreground = Drawable.createFromXml(
+                resources,
+                resources.getXml(R.drawable.gradient_light_to_transparent_from_bottom)
```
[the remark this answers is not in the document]
SanJSp: it is of type xml

app/src/main/res/layout/fragment_video_preview.xml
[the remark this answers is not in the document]
tobiasrohloff: Fine for me, Android 5 is from 2014.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -188,4 +224,52 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+
+    private fun displayCollapsedDescription() {
+        textDescription.visibility = View.VISIBLE
+        showDescriptionButton.visibility = View.VISIBLE
+        textDescriptionTitle.visibility = View.VISIBLE
+        textDescription.ellipsize = TextUtils.TruncateAt.MARQUEE
+        textDescription.maxLines = 3
+        showDescriptionButton.text = getString(R.string.show_more)
+
+        if (textDescription.text.length < 115) {
```
tobiasrohloff, line 248: That's the magic number, right? Actually you want to know if your text was ellipsized. Try out things like that: https://stackoverflow.com/a/4008755/4870048

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -113,6 +131,24 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
 
         textTitle.text = item.title
 
+        if (VideoDescriptionDelegate.isVideoSummaryAvailable(video.summary)) {
+            textDescription.setTypeface(textDescription.typeface, Typeface.NORMAL)
+            textDescription.setMarkdownText(video.summary)
+            displayCollapsedDescription()
+
+            showDescriptionButton.setOnClickListener {
+                if (showDescriptionButton.text == getString(R.string.show_more)) {
```
tobiasrohloff, line 140: I don't like the state handling based on the button label.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -188,4 +224,52 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+        textDescription.maxLines = 3
+        showDescriptionButton.text = getString(R.string.show_more)
+
+        if (textDescription.text.length < 115) {
+            showDescriptionButton.visibility = View.GONE
+            if (Feature.FOREGROUNDS) {
+                textDescription.foreground = null
+            }
+        } else if (Feature.FOREGROUNDS) {
+            textDescription.foreground = Drawable.createFromXml(
```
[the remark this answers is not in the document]
tobiasrohloff: Isn't it possible to get the `Drawable` with `ContextCompat.getDrawable(getActivity(), R.drawable.name)`?

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -188,4 +224,52 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+                resources,
+                resources.getXml(R.drawable.gradient_light_to_transparent_from_bottom)
+            )
+        }
+    }
+
+    private fun displayFullDescription() {
+        textDescription.ellipsize = null
+        textDescription.maxLines = 1000
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
```
tobiasrohloff, line 264: Isn't that the `Feature.FOREGROUNDS` check?

app/src/main/java/de/xikolo/config/Feature.kt
```diff
@@ -16,10 +16,11 @@ object Feature {
 
     @JvmField
     val PIP =
-        Build.VERSION.SDK_INT >= 26 && App.instance.packageManager
+        SDK_INT >= 26 && App.instance.packageManager
             .hasSystemFeature(PackageManager.FEATURE_PICTURE_IN_PICTURE)
 
-    val SHORTCUTS = Build.VERSION.SDK_INT >= 25
+    val SHORTCUTS = SDK_INT >= 25
+    val FOREGROUNDS = SDK_INT >= 23
```
tobiasrohloff, line 23: Maybe name it a bit more specific like `VIEW_FOREGROUNDS`.

tobiasrohloff (changes requested):
I think Screenshot 3 is an unfavorable example since the third line is an empty one. Can you please post a better example? (But don't worry, we don't start to optimize the text here for these edge cases.)

mathebox:
How about always showing the full description (if there is one)?

tobiasrohloff:
@mathebox we decided against that. One reason was to not scroll forever to find the download buttons.

app/src/main/res/layout/fragment_video_preview.xml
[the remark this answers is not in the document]
bengelhaupt: Yep, that's because pre-23 `android:foregound` could only be applied to `FrameLayout`. If you want to fix this, you could wrap that code inside a FrameLayout or add a simple `View` with the fade as its background atop the TextView.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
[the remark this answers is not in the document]
bengelhaupt: I think they mean an Integer that is annotated with `@DrawableRes` (R.drawable.....)

app/src/main/java/de/xikolo/config/Feature.kt
bengelhaupt: I would build the layout with a `FrameLayout`, so `android:foreground` works API21+ and these "Feature" checks are not necessary.
See https://stackoverflow.com/questions/52329249/why-doesnt-androidforeground-attribute-work

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -44,6 +50,18 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+    lateinit var textSubtitles: TextView
+
+    @BindView(R.id.headerTextPreviewDescription)
+    lateinit var textDescriptionTitle: TextView
+
+    @BindView(R.id.textPreviewDescription)
+    lateinit var textDescription: TextView
+
+    @BindView(R.id.showDescriptionButton)
+    lateinit var showDescriptionButton: TextView
```
bengelhaupt, line 63: This should be a button

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -113,6 +131,24 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
 
         textTitle.text = item.title
 
+        if (VideoDescriptionDelegate.isVideoSummaryAvailable(video.summary)) {
```
bengelhaupt, line 134: That's not how delegates were meant to be used. They should only be used by ViewModels and calls be wrapped.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -188,4 +224,52 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
...
+            textSubtitles.visibility = View.VISIBLE
+        } else {
+            textSubtitles.visibility = View.GONE
+        }
+    }
+
+    private fun displayCollapsedDescription() {
+        textDescription.visibility = View.VISIBLE
+        showDescriptionButton.visibility = View.VISIBLE
+        textDescriptionTitle.visibility = View.VISIBLE
```
bengelhaupt, line 243: Wrap those inside a layout container and only hide/show the whole container

app/src/main/java/de/xikolo/viewmodels/shared/VideoDescriptionDelegate.kt
```diff
@@ -0,0 +1,24 @@
+package de.xikolo.viewmodels.shared
+
+import de.xikolo.models.VideoSubtitles
+import de.xikolo.utils.LanguageUtil
+import io.realm.RealmList
+
+object VideoDescriptionDelegate {
```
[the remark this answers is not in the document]
bengelhaupt: Yes, please move these to the `Video` class

app/src/main/res/drawable/gradient_light_to_transparent_from_bottom.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
+
+    <gradient
+        android:angle="270"
+        android:centerColor="#4DFFFFFF"
```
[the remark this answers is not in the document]
bengelhaupt: Also please only declare colors in the color resources

app/src/main/res/layout/fragment_video_preview.xml
```diff
@@ -25,18 +25,63 @@
...
+                android:maxLines="3"
+                android:textSize="14sp" />
+
+            <TextView
+                android:id="@+id/showDescriptionButton"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_below="@+id/textPreviewDescription"
+                android:layout_alignParentEnd="true"
+                android:paddingTop="8dp"
```
bengelhaupt, line 73: I personally would use a smaller margin (not padding, because it makes the view bigger) to the top. I feel the distance between the button and view is quite big

SanJSp:
@tobiasrohloff here is an alternative video item with better screenshots. Will work on the feedback, thank you 😊

mathebox:
You could also consider omitting the label for description ("Beschreibung"), especially as teachers don't always provide a description but some additional information.

tobiasrohloff:
>You could also consider omitting the label for description ("Beschreibung"), especially as teachers don't always provide a description but some additional information.

@SanJSp please check the layout in the actual video player view (example in #252). We don't use a headline there as well.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -113,6 +131,24 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
 
         textTitle.text = item.title
 
+        if (VideoDescriptionDelegate.isVideoSummaryAvailable(video.summary)) {
+            textDescription.setTypeface(textDescription.typeface, Typeface.NORMAL)
```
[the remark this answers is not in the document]
SanJSp: I've copied the code from `VideoDescriptionFragment`, although leaving it out doesn't change anything.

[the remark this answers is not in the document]
tobiasrohloff: @bengelhaupt any idea why we have it in `VideoDescriptionFragment`?

app/src/main/res/layout/fragment_video_preview.xml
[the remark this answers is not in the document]
SanJSp: I will implement the FrameLayout solution 😃 it will also get rid of the Feature.FOREGROUND checks

SanJSp:
The current look of the solution. The added margin on the expanded text is because of newlines in the summary.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
```diff
@@ -44,6 +49,18 @@ class VideoPreviewFragment : ViewModelFragment<VideoPreviewViewModel>() {
     @BindView(R.id.textTitle)
     lateinit var textTitle: TextView
 
+    @BindView(R.id.textSubtitles)
```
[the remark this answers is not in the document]
bengelhaupt: I get those warnings as well. AutoBundle is not well maintained and I guess we will need to migrate to [View Binding](https://developer.android.com/topic/libraries/view-binding) soon.

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,129 @@
...
+        var audioSize = 0
+
+        @field:Json(name = "transcript_url")
+        var transcriptUrl: String? = null
+
+        @field:Json(name = "transcript_size")
+        var transcriptSize = 0
+
+        @field:Json(name = "thumbnail_url")
+        var thumbnailUrl: String? = null
```
bengelhaupt, line 93: Missing newlines

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,129 @@
...
+            return singleStream as VideoStream
+        }
+        if (lecturerStream != null && (lecturerStream!!.hdUrl != null || lecturerStream!!.sdUrl != null)) {
+            return lecturerStream as VideoStream
+        } else return slidesStream!!
+    }
+
+    @JsonApi(type = "videos")
+    class JsonModel : Resource(), RealmAdapter<Video> {
+        var summary: String? = null
```
bengelhaupt, line 62: Missing newlines

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    var transcriptUrl: String? = null
+
+    var transcriptSize = 0
+
+    var thumbnailUrl: String? = null
+
+    var subtitles = RealmList<VideoSubtitles>()
+
+    fun getItem(): Item? {
```
bengelhaupt, line 44: Refactor to `val`. Makes the other changes to `getItem()` unnecessary

bengelhaupt (changes requested):
Good job!
I noticed some small things (Course Successful Remote Teamwork, Section 1):

This seems to have two lines of description but there is "no more" There would have to be two line breaks at the end to render this, although when expanding it does not seem so. Expected behavior: with <= 3 Lines no "Show More" button is shown.


Occasionally the descriptio container is weirdly undercut. There is no clear way to reproduce it other than scrolling and expanding, collapsing and refreshing a few times.


I think this also has to do with it. When collapsed, not the top of the description is shown.

app/src/main/java/de/xikolo/controllers/section/VideoPreviewFragment.kt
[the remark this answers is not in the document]
tobiasrohloff: Yeah, View Binding looks promising but also like a hell of a refactoring. Ignore these for now. Maybe we can disable these warnings in our linter config (any volunteers?).

BTW, this is a ButterKnife annotation, not AutoBundle. But I would love to replace both of course.

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,131 @@
...
+
+    @Ignore
+    var item: Item? = findForContent(id)
+
+    // local field
+    var progress = 0
+
+    val streamToPlay: VideoStream
+        get() {
+            return if (singleStream != null && (singleStream!!.hdUrl != null || singleStream!!.sdUrl != null)) {
```
bengelhaupt, line 53: The `singleStream!!` call is still unsafe. Consider using `singleStream?.hdUrl != null`.

app/src/main/java/de/xikolo/models/DownloadAsset.kt
```diff
@@ -94,7 +94,7 @@ object DownloadAsset {
             }
 
             class VideoSD(item: de.xikolo.models.Item, val video: Video) :
-                Item(video.streamToPlay.sdUrl, "video_sd_${item.id}.mp4", item) {
+                Item(video.streamToPlay?.sdUrl, "video_sd_${item.id}.mp4", item) {
```
bengelhaupt, line 97: Now that I see that the access was safe before, I think it might be better to make `streamToPlay` of type `VideoStream?` again and make an unsafe `!!` call only where there is no other possibility. Sorry for my misjudgement

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,131 @@
...
+    var transcriptUrl: String? = null
+
+    var transcriptSize: Int = 0
+
+    var thumbnailUrl: String? = null
+
+    var subtitles = RealmList<VideoSubtitles>()
+
+    @Ignore
+    var item: Item? = findForContent(id)
```
bengelhaupt, line 46: Should have a getter. This way the field is initialized on class init where `id` might be `null` and any call to `item` always returns `null`.

app/src/main/res/layout/fragment_video_preview.xml
```diff
@@ -25,18 +25,54 @@
...
+                android:layout_below="@+id/textSubtitles"
+                android:layout_marginTop="8dp"
+                android:foreground="@drawable/gradient_light_to_transparent_from_bottom">
+
+                <TextView
+                    android:id="@+id/textPreviewDescription"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:ellipsize="end"
+                    android:maxLines="3"
```
bengelhaupt, line 56: Does this need to be set statically or is it possible to set it programmatically to our constant?

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    fun getSubtitlesAsString(): String {
+        return subtitles.map {
+            LanguageUtil.toLocaleName(it.language)
+        }.joinToString(", ", "")
+    }
+
+    val isSummaryAvailable: Boolean
```
bengelhaupt, line 124: So I guess `val`s with a getter are not being considered by Realm? Please move `item` and `streamToPlay` also down to these "helper" fields

bengelhaupt: The `@Ignore` tag is then not needed

bengelhaupt (changes requested):
Were you able to reproduce / fix the cut video description problem?

SanJSp:
I've been able to reproduce the problem, but only twice in about thirty tries. Maybe it's a bug because I've not been able to reproduce it with other video descriptions.

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,132 @@
...
+            video.thumbnailUrl = thumbnailUrl
+            video.subtitles.addAll(subtitles!!)
+            return video
+        }
+    }
+
+    val streamToPlay: VideoStream?
+        get() {
+            return if (singleStream != null && (singleStream?.hdUrl != null || singleStream?.sdUrl != null)) {
+                singleStream as VideoStream
```
bengelhaupt, line 108: Remove this cast. Casting to nun-null `VideoStream` is essentially the same as using `!!`.

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,132 @@
...
+            } else if (lecturerStream != null && (lecturerStream?.hdUrl != null || lecturerStream?.sdUrl != null)) {
+                lecturerStream as VideoStream
```
bengelhaupt, line 110: Same here

app/src/main/java/de/xikolo/models/Video.kt
```diff
@@ -0,0 +1,132 @@
...
+            } else slidesStream as VideoStream
```
bengelhaupt, line 111: Same here

bengelhaupt (changes requested):

bengelhaupt (approved):

tobiasrohloff (approved):
Thank you!

Issue #269, opened 2020-08-17
Enroll deeplink does not show CourseActivity in `enrolled` state


bengelhaupt:
When clicking on the enroll button on the web and being led to the app, the `CourseActivity` is not started in the enrolled mode and only shows description and certificate tabs.

This is caused by not having an active observer after initialization. But when adding such an observer and the user is led into the app by such an enrollment deeplink, the activity would need to be restarted as there was a bug with our viewpager.

PR #285, opened 2020-10-13
Reload course model on CourseActivity refresh

### LINKED ISSUE #284: Enrolled Courses Aren't Accessible After Start
We got user reports that courses that didn't start yet and a user already enrolled in it, didn't become accessible after the course start date was passed. The notice "Course has not been started yet" was still shown.

I was able to reproduce this but it can be tricky to duplicate such a state with production data. Maybe manipulate the incoming data or use dummy data, like the tests.

I experienced the issue for the following courses, but they obviously have started in the meantime:
- https://open.hpi.de/courses/kieinstieg2020/
- https://open.hpi.de/courses/learningtheory2020/

Clearing the app memory has solved the problem as a workaround.

bengelhaupt:
I was trying to reproduce the bug of #284 but couldn't, because it had already been fixed in #279. So I guess the bugs only occurred in app version 3.5. #279 states `CourseActivity: Our workaround was to observe the course only once and to ignore changes resulting e.g. from enrollments (resulting in bugs like #269).`. This also included changes in course area states such as course accessibility, because no observer would be active in the activity, resulting in not setting it up again when the user refreshed (and possibly always operating on outdated data that is returned on the first observe). #279 introduced such an observer which sets up the course when the database model changes. When only the tabs "Description" and "Certificates" are displayed, both of their fragments fetch the course model on refresh, which means any refresh in them also refreshes the activity (which is convenient). Other fragments such as "Learnings" do not do this. As an enhancement I made the activity also react to clicking "Refresh" in the menu.

I tested the behavior by manually altering the value of `accessible` in the `Course.JsonModel` in the `GetCourseJob` and applying changes to the activity without restarting the app. It rendered following behavior:

Fixes #284

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
+++ b/app/src/main/java/de/xikolo/controllers/course/CourseActivity.kt
@@ -265,7 +265,7 @@ class CourseActivity : ViewModelActivity<CourseViewModel>(), UnenrollDialog.List
 
     override fun onOptionsItemSelected(item: MenuItem): Boolean {
         when (item.itemId) {
-            android.R.id.home    -> {
+            android.R.id.home -> {
                 if (supportFragmentManager.backStackEntryCount > 0) {
                     NavUtils.navigateUpFromSameTask(this)
                 } else {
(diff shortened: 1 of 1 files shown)


tobiasrohloff (approved):
Alright, thanks! I will monitor if that issue still exists in 3.6 or not.

tobiasrohloff:
If you rebase to the latest master, Danger should also work again here!

PR #288, opened 2020-10-20
Add intent query declaration for Android 11

### LINKED ISSUE #286: Cannot Open LTI Exercises
Currently, it seems that opening LTI exercises don't work anymore. A toast message states that "an error occurred". The expected behavior is that a browser window is opened with the LTI exercise for the current user.

I tested it with the emulator and my Pixel 3, both run Android 11.

It could be a regression from the last deep link fixes.

Also, it should work with external LTI tools and inline LTI tools, which are rendered in an iframe for the web app. We don't differentiate the behavior on mobile here, both types are opened externally. I already tested the API responses and they work fine for both (and the API also shows the same behavior for both).

An external LTI exercise:
https://open.hpi.de/courses/ruby2018/items/3ILkiD5AIgbeX11ix3bQz0

An inline LTI exercise (hidden and invite-only test course):
https://open.hpi.de/courses/mwtest/items/6LETxwi26TLse5yjZ7hNYo

bengelhaupt:
Fixed as per https://developer.android.com/training/basics/intents/package-visibility#intent-signature

Fixes #286

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,6 +15,10 @@
             <action android:name="android.intent.action.VIEW" />
             <data android:scheme="http" />
         </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="https" />
+        </intent>
     </queries>
 
     <application
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -10,6 +10,13 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
 
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="http" />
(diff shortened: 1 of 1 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -10,6 +10,13 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
 
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="http" />
```
tobiasrohloff, line 16: Does that work for `https` links as well?

tobiasrohloff (approved):

app/src/main/AndroidManifest.xml
bengelhaupt: We do use that schema to create a "fakeIntent" for `http://someurl.com` to query browser apps. I will add "https" too just to have the set complete.

tobiasrohloff:
I tested both items with Android 10 and 11 on the emulator and everything works as expected again, thanks! :)

PR #293, opened 2020-10-28
Allow back navigation in Discussions WebView

bengelhaupt:
When navigating the discussions tab in the `CourseActivity` and going back, instead of leading the user to the last visited page, the whole activity closes. It annoys myself and I also heard some user feedback on this. 
I included back handling for our `WebViewFragment` in general and enabled it only for discussuions for now (because e.g. for Recaps, no going back should be possible I guess.)


tobiasrohloff (approved):
Lgtm, but I'm wondering since when this doesn't work anymore. It was definitely possible in the past.

PR #294, opened 2020-10-28
Fix vertical scrolling in nested WebViews

### LINKED ISSUE #291: Scrolling in WebViews leads to ViewPager page swipes
To reproduce that behavior, open a course forum and scroll down. This leads to page swipes to the right or left (the other tabs). Probably an issue between ViewPager2 and our NestedScrollWebView.

bengelhaupt:
I found a fix for #291 by checking whether the current move event is primarily happening on the y-axis (delta-y > delta-x && something to scroll) and calling `requestDisallowInterceptTouchEvent(true)` to prevent the ViewPager from intercepting it and attempting a tab switch.

I tested it on my phone and emulator and I am satisfied with the behavior. 
Please check the behavior as well when reviewing.

Fixes #291

### DIFF
--- a/app/src/main/java/de/xikolo/views/NestedScrollWebView.kt
+++ b/app/src/main/java/de/xikolo/views/NestedScrollWebView.kt
@@ -24,6 +24,7 @@ import android.webkit.WebView
 import androidx.core.view.NestedScrollingChild
 import androidx.core.view.NestedScrollingChildHelper
 import androidx.core.view.ViewCompat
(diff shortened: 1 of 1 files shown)


SanJSp:
Tested it and it works out fine :)

SanJSp (approved):

tobiasrohloff:
That works pretty well, thank you! 😘

tobiasrohloff (approved):

PR #302, opened 2021-03-07
Subtitle support for Cast

### LINKED ISSUE #301: Subtitle Support for Cast
When streaming a video with a Google Cast device it should be possible to select a subtitle track.

https://developers.google.com/cast/docs/android_sender/media_tracks

bengelhaupt:
Adds a button for subtitle track selection.
To make the former 3-button layout (30s backward, play/pause, mute) symmetric, the 30s formward button has been added as well.
The layout now is (subtitles, 30s backward, play/pause, 30s forward, mute).

TODO:
- ~~testing~~
- ~~hiding button when no subtitles available?~~

Fixes #301

### DIFF
--- a/app/src/main/java/de/xikolo/utils/extensions/CastExtensions.kt
+++ b/app/src/main/java/de/xikolo/utils/extensions/CastExtensions.kt
@@ -7,6 +7,8 @@ import android.net.Uri
 import com.google.android.gms.cast.MediaInfo
 import com.google.android.gms.cast.MediaLoadOptions
 import com.google.android.gms.cast.MediaMetadata
+import com.google.android.gms.cast.MediaTrack
+import com.google.android.gms.cast.TextTrackStyle
 import com.google.android.gms.cast.framework.CastContext
 import com.google.android.gms.cast.framework.CastState
 import com.google.android.gms.cast.framework.media.RemoteMediaClient
@@ -15,6 +17,7 @@ import com.google.android.gms.common.images.WebImage
 import com.google.firebase.crashlytics.FirebaseCrashlytics
 import de.xikolo.controllers.cast.CastActivity
 import de.xikolo.models.Video
+import de.xikolo.utils.LanguageUtil
 
 val <T : Context> T.isCastConnected: Boolean
     get() {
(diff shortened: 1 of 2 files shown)


tobiasrohloff:
>hiding button when no subtitles available?

Yes, please. :)

bengelhaupt:
The current client-side implementation works, but the subtitle resources lack CORS headers.
I added the headers `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Headers: Content-Type, Accept-Encoding, Range` in a test environment (https://www.bengelhaupt.com/hlstest/raw/subs) and it worked.



The UI automatically greys out the closed captions button when there are none available, which is the best solution I think.


The actual selection dialog provided by the framework looks a bit dated but does its job.

tobiasrohloff:
CORS headers again, yay. They are needed here independently from the whole HLS thing?

bengelhaupt:
> CORS headers again, yay. They are needed here independently from the whole HLS thing?

Yes it seems so, because the Cast Receiver is web-based and obeys the whole CORS thing with its rules.

tobiasrohloff:
@bengelhaupt I created a backend ticket for that. While I was specifying that, two questions came to my mind:
- Is the `Access-Control-Allow-Headers` required? Or does it already works with `Access-Control-Allow-Origin` only?
- Do the headers have to be returned only for `GET` requests, or also for `OPTIONS`?

Can you please test these two things in all four combinations? This could also help us understand what exactly is required for the HLS playlists.

bengelhaupt:
> @bengelhaupt I created a backend ticket for that. While I was specifying that, two questions came to my mind:
> 
> * Is the `Access-Control-Allow-Headers` required? Or does it already works with `Access-Control-Allow-Origin` only?
> * Do the headers have to be returned only for `GET` requests, or also for `OPTIONS`?
> 
> Can you please test these two things in all four combinations? This could also help us understand what exactly is required for the HLS playlists.

It already works when just the `Access-Control-Allow-Origin` header on the GET request is present. OPTIONS requests were not sent at all.

|  | ACAO only | ACAO+ACAH |
| --- | --- | --- |
| GET | ✔️ | ✔️ |
| OPTIONS | not sent | not sent |

tobiasrohloff:
@bengelhaupt the ACAO header for `GET` requests is live. Can you please test it again?

bengelhaupt:
@tobiasrohloff Works perfectly 👍

PR #303, opened 2021-03-22
Fix intent chooser component querying

### LINKED ISSUE #300: Cannot Open LTI Exercises, again
We have issues with LTI exercises again (older bug: #286).

It can be reproduced, for example, with this item:
https://open.hpi.de/courses/pythonjunior2020/items/rHiz0imat9JxB5Ih7K1PQ

bengelhaupt:
This should now ultimately fix our intent query issue.

Our main requirement was to hide the own app in Intent Choosers. Our approach was to explicitly specify the apps to show in the Chooser using the flag `Intent.EXTRA_INITIAL_INTENTS`, requiring the scan of all apps on the device (which does not function properly on Android 11). The soultion is to now exclude the components from our own app explicitly in the Chooser via the flag `Intent.EXTRA_EXCLUDE_COMPONENTS`. As this works only from N on, we use the old working logic for sub-N.
The only limitation now is that we cannot hide other apps than our own, but we did not use that functionality anyway.

For testing, better uninstall the production app (or set its intent opening settings to default).

Fixes #300

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -10,21 +10,6 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
 
-    <queries>
-        <intent>
-            <action android:name="android.intent.action.VIEW" />
-            <data android:scheme="http" />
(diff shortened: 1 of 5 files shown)


tobiasrohloff (approved):
Sounds very reasonable! 👍 Code looks fine as well.

PR #306, opened 2021-04-21
Update dependencies

tobiasrohloff:


bengelhaupt:
I tested this, especially video playback (because of ExoPlayer update), and everything seems fine.

PR #308, opened 2021-04-21
Fix subtitle text in video settings

bengelhaupt:
There was a bug introduced in https://github.com/openHPI/xikolo-android/commit/b56290c7ca8a00c9f9b9791fee2c2e32d9b81626 that appended a `null` to the subtitles text when no subtitles were selected. This has been fixed by adding proper parentheses.


tobiasrohloff:
Thx!

PR #311, opened 2021-06-13
Fix RequestManagerRetriever.java line 348

bengelhaupt:
Sometimes when `cancel()` is called the Activity and thus the `context` has been destroyed. Caching it fixes the problem.

See https://github.com/noties/Markwon/issues/259


tobiasrohloff:
Thx!

PR #312, opened 2021-06-13
Fix VideoPreviewFragment.kt line 267

bengelhaupt:
The `post()` on the TextView could cause the block to be executed when the Fragment had already been detached.


tobiasrohloff:
Thx!

PR #318, opened 2021-08-29
Fix ProgressListAdapter.kt line 143

bengelhaupt:
There were some malformed string formatting rules in the chinese string resources that caused a crash.

See [https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/972342597b693c0f2fd5df35f31f1c73](https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/972342597b693c0f2fd5df35f31f1c73)


tobiasrohloff:
Good catch, thanks!

PR #319, opened 2021-08-29
Fix unnecessary blank space in collapsing toolbar activities

### LINKED ISSUE #307: Fix unnecessary blank space in collapsing toolbar activities
While going through the app I noticed some blank space that seemed unusual and that I have not seen before. It appears above the RecyclerViews inside of collapsing toolbar activities. When scrolling (so when the toolbar is collapsed) the content scrolls underneath it, and also the last card in the RecyclerView is cut off slightly (non-critical, the buttons are clickable).

I tracked the appearance down to our Android 11 commit https://github.com/openHPI/xikolo-android/commit/7102b805f5d08103a33b87f34ad8e6b439a28673.

It could be caused by a dependency update or simply the targetSdk change to Android 11, but anyways it is probably tedious to resolve.

bengelhaupt:
I did a trial-and-error search for the bug and found out that a single `android:fitsSystemWindows="true"` had to be removed.

Fixes #307

### DIFF
--- a/app/src/main/res/layout/activity_blank_collapsing.xml
+++ b/app/src/main/res/layout/activity_blank_collapsing.xml
@@ -71,7 +71,6 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:background="@color/background"
-            android:fitsSystemWindows="true"
             app:layout_behavior="@string/appbar_scrolling_view_behavior" />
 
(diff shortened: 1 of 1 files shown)


tobiasrohloff:
Interesting, I don't remember why this was added. Thanks for the investigation!

PR #320, opened 2021-08-29
Remove retainInstance from WebViewFragment

### LINKED ISSUE #292: llegalStateException for AutoBundles: required fields not found in the bundle
Required fields are often empty for different AutoBundles. See the different errors [here](https://console.firebase.google.com/project/openwho-firebase/crashlytics/app/android:de.xikolo.openwho/issues/b5f0e64ce6e6ebb2fd19498172096942?time=last-thirty-days&versions=3.6%20(60)).

bengelhaupt:
[Fragment.setRetainInstance()](https://developer.android.com/reference/androidx/fragment/app/Fragment.html#setRetainInstance(boolean)) has been used to retain the state of our `WebViewFragment`, e.g. to preserve user input when the screen has been rotated. It is deprecated and I highly assume that it has something to do with #292 because of its weird state handling. I rewrote the logic to use the instance-state Bundles.

Maybe fixes #292

PS. I accidentally pushed to master, but I reverted the commit. Sorry 🤦‍♂️

### DIFF
--- a/app/src/main/java/de/xikolo/controllers/webview/WebViewFragment.kt
+++ b/app/src/main/java/de/xikolo/controllers/webview/WebViewFragment.kt
@@ -1,17 +1,11 @@
 package de.xikolo.controllers.webview
 
 import android.content.Intent
-import android.content.MutableContextWrapper
 import android.net.Uri
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


tobiasrohloff:
Thanks!

